PR #18, opened 2024-05-01
Improved lifi integration

krzkaczor:
- [x] set lower default max slippage
- [x] fix max slippage parameter name sent to lifi api
- [ ] add maxPriceImpact parameter when calling lifi api


packages/app/src/domain/exchanges/lifi/meta/index.ts
```diff
@@ -15,9 +17,14 @@ export interface EvaluateParams {
   toToken: CheckedAddress
 }
 
+export interface EvaluateResult {
```
yivlad, line 20: EvaluationResult

packages/app/src/domain/exchanges/lifi/query.ts
```diff
@@ -74,7 +78,14 @@ export function fetchLiFiTxData({
           },
         }
       } else {
-        const response = await client.getReverseQuote({ fromToken, toToken, amount, meta, maxSlippage })
+        const response = await client.getReverseQuote({
+          fromToken,
+          toToken,
+          amount,
+          maxSlippage,
```
yivlad, line 85: I guess it should pass maxPriceImpact as well?

packages/app/src/domain/exchanges/lifi/lifi.ts
```diff
@@ -98,7 +107,13 @@ export class LiFi {
     return url
   }
 
-  private buildReverseQuoteRequestOptions({ fromToken, toToken, amount, maxSlippage }: GetQuoteOptions): RequestInit {
+  private buildReverseQuoteRequestOptions({
```
oskarvu, line 110: It takes `maxPriceImpact` as a param, but later, `getReverseQuote` doesn't have it in options.

packages/app/src/domain/exchanges/lifi/meta/index.ts
krzkaczor: I just follow naming pattern of "methodname" + "Result"

packages/app/src/domain/state/actions-settings.ts
```diff
@@ -16,13 +16,15 @@ export interface ActionsSettingsSlice {
   actionsSettings: ActionsSettings
 }
 
+const defaultExchangeMaxSlippage = Percentage(0.0005)
```
oskarvu, line 19: Will it work for routes that can't use sc directly? USDC -> SDAI for example.

packages/app/src/domain/exchanges/lifi/query.ts
krzkaczor: It's part of overrides (next line). and btw. Support for maxPriceImpact def needs test but for now i tested it manually.

yivlad (approved):

packages/app/src/domain/exchanges/lifi/lifi.ts
```diff
@@ -70,14 +74,23 @@ export class LiFi {
       toToken,
       integrator: meta.integratorKey,
```
yivlad, line 75: Can you please remind and leave a comment why we have integrator and fee in buildQuote but not in buildReverseQuote?

oskarvu (approved):

packages/app/src/domain/exchanges/lifi/lifi.ts
krzkaczor: This should be probably fixed but it's unrelated to what we are dealing with now.

PR #24, opened 2024-05-08
Replace tooltips with popovers on touch screens

yivlad:


packages/app/src/ui/atoms/tooltip/Tooltip.tsx
```diff
@@ -1,13 +1,15 @@
+import * as PopoverPrimitive from '@radix-ui/react-popover'
 import * as TooltipPrimitive from '@radix-ui/react-tooltip'
 import * as React from 'react'
 
 import { cn } from '@/ui/utils/style'
 
 const TooltipProvider = TooltipPrimitive.Provider
```
oskarvu, line 7: Should provider also be swapped in case of touch screen?

oskarvu (approved):

PR #32, opened 2024-05-10
Add quote price difference warning

yivlad:


packages/app/src/config/feature-flags/appConfig.default.ts
```diff
@@ -14,7 +14,7 @@ export function getDefaultAppConfig(): AppConfig {
 
       mintBalances: {
         etherAmt: NormalizedUnitNumber(10_000),
-        tokenAmt: NormalizedUnitNumber(10_000),
```
krzkaczor, line 17: do we need to change it?

packages/app/src/features/dialogs/savings/deposit/logic/useRiskAcknowledgement.ts
```diff
@@ -0,0 +1,50 @@
...
+  swapInfo: SwapInfo
+  marketInfo: MarketInfo
+  potParams: PotParams
+  inputValues: DialogFormNormalizedData
+}
+
+export interface RiskAcknowledgementInfo {
+  acknowledged: boolean
+  isRiskAcknowledgedOrNotRequired: boolean
+  onStatusChanged: (acknowledged: boolean) => void
```
oskarvu, line 21: nit: `onStatusChange`

packages/app/src/features/dialogs/savings/withdraw/logic/generateTransactionWarning.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const fromAmountDAI = convertSharesToDai({
+    shares: sDAI.fromBaseUnit(swapInfo.data.estimate.fromAmount),
+    timestamp,
+    potParams,
+  })
+  const toAmountMinDAI = inputToken.fromBaseUnit(swapInfo.data.estimate.toAmountMin)
+
+  const discrepancy = NormalizedUnitNumber(fromAmountDAI.minus(toAmountMinDAI))
+  if (discrepancy.gte(WARNING_DISCREPANCY_THRESHOLD)) {
```
oskarvu, line 48: I wonder if we should use `DAI` here? Didn't we agree, that maybe difference should be denominated in USD? It will be more 'universal' (for example when withdrawing to USDC/USDT). And same with deposit I guess.

packages/app/src/features/dialogs/savings/withdraw/logic/useRiskAcknowledgement.ts
oskarvu, line 1: Isn't this file identical to the hook used at savings deposit?

packages/app/src/features/dialogs/savings/withdraw/logic/generateTransactionWarning.ts
yivlad: I think we want to denominate everything in DAI

packages/app/src/features/dialogs/savings/deposit/logic/generateWarning.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface GenerateWarningArgs {
+  swapInfo: SwapInfo
+  inputValues: DialogFormNormalizedData
+  marketInfo: MarketInfo
+  potParams: PotParams
+  timestamp: number
+}
+
+export interface GenerateWarningResults {
```
krzkaczor, line 20: Why not make this return even simpler type:
```typescript
shouldDisplayWarning: boolean
discrepency: NormalizedNumber
```

and then make RiskAcknowledgment component take children with the exact warning message that would be simply hardcoded in the parent controller? This way we totally unbound visual aspect like formatting, of the error from logic.

krzkaczor (approved):

oskarvu (approved):

PR #35, opened 2024-05-13
Replay integration

krzkaczor:
Closes FRO-676

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,8 +115,10 @@ jobs:
           cache: 'pnpm'
       - run: pnpm install
 
-      # - run: pnpm exec playwright install --with-deps chromium
-      #   working-directory: ./packages/app
+      - run: |
+          pnpx @replayio/playwright install
+        working-directory: ./packages/app
+
       - run: pnpm run test-e2e
         working-directory: ./packages/app
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,8 +115,8 @@ jobs:
           cache: 'pnpm'
       - run: pnpm install
 
-      - run: pnpm exec playwright install --with-deps chromium replay-chromium
-        working-directory: ./packages/app
+      # - run: pnpm exec playwright install --with-deps chromium
+      #   working-directory: ./packages/app
       - run: pnpm run test-e2e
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -115,27 +115,15 @@ jobs:
...
 
-      - name: TMP fix MS repos
-        run: |
-          sudo rm /etc/apt/sources.list.d/microsoft-prod.list
-          sudo apt-get update
-
-      - run: pnpm exec playwright install --with-deps chromium
-        working-directory: ./packages/app
+      # - run: pnpm exec playwright install --with-deps chromium
+      #   working-directory: ./packages/app
```
bvaughn, line 119: Try?
```suggestion
      # install the native deps of Chromium itself which happens to be the same as the ones of Replay Chromium
      - run: npx playwright install-deps chromium
        if: steps.playwright-cache.outputs.cache-hit != 'true'
```

Andarist: This might be a good thing to add regardless but I think it won't help. The CI action failed to locate our executable and this is used to install our dependencies - not our browser itself. 

I think that *maybe* our postinstall script failed to do its thing. I can't find any configuration here that would prevent it from running but 🤷 

Maybe in addition to what Brian's suggested you could also run this as "postinstall" yourself? `replayio-playwright install`

bvaughn: Ah, maybe you're right.

> Maybe in addition to what Brian's suggested you could also run this as "postinstall" yourself? replayio-playwright install

I think you mean `@replayio/playwright install` ?

(That is also what the Devtools project runs)

Andarist: > I think you mean @replayio/playwright install ?

This probably works with `npx`:
https://github.com/replayio/devtools/blob/0518bec9b45d9b136359452b2af0c8dcc431eb44/.github/workflows/test.yml#L100-L101

but if you are looking for an executable added by `@replayio/playwright` that's actually `replayio-playwright` (a lengthy one 😬 ):
https://github.com/replayio/replay-cli/blob/e3d7e8f8190bbe51cde55815db5421e7f3cab9c0/packages/playwright/package.json#L6-L8

My guess is that `npx` only works like this when u have a single bin but you can register multiple and name them however you want.

PR #48, opened 2024-05-16
Spark airdrop panel

paszkowskiDamian:
- Airdrop information added to market details pages
- New molecule component added for creating Spark informations
- Extracted docs link to atom component

image

image


packages/app/src/ui/molecules/spark-info-panel/SparkInfoPanel.tsx
```diff
@@ -0,0 +1,30 @@
...
+      className={cn(
+        'mt-6 py-4',
+        'col-start-1 col-end-[-1] grid grid-cols-[auto_1fr] grid-rows-[auto_auto]',
+        'bg-spark/10 border-none',
+      )}
+    >
+      <Panel.Title variant="h4" className={cn('col-span-1 content-end')}>
+        {title}
+      </Panel.Title>
+      <Panel.Content className={cn('row-span-2 row-start-1 m-4 content-center')}>
```
yivlad, line 22: I'd put everything inside one `<Panel.Content>` and make it a grid

packages/app/src/features/market-details/components/status-panel/BorrowStatusPanel.tsx
```diff
@@ -78,6 +86,18 @@ export function BorrowStatusPanel({
         <div className="col-span-3 mt-6 sm:mt-10">
           <InterestYieldChart {...chartProps} />
         </div>
+        {airdropEligible && (
+          <SparkInfoPanel
+            title={<>Eligible for 24M Spark Airdrop</>}
+            content={
+              <>
+                {getNativeTokenSymbolIfWrapped(chainId, token.symbol)} borrowers will be eligible for a future ⚡ SPK
```
yivlad, line 94: I think we should do i`getNativeTokenSymbolIfWrapped(chainId, token.symbol)` in `useMarketDetails` hook and prop drill to the panel. Sometimes one market details page is displaying info for 2 markets -  an example is DAI market details, which displays info from DAI and sDAI market. Calculating the airdrop token in master hook will allow to easily account for this as well.

packages/app/src/ui/molecules/spark-info-panel/SparkInfoPanel.tsx
oskarvu, line 1: - I would not use `Panel` atom here and just create custom jsx.
- I think it doesn't need to go to the molecules, but rather to `market-details/components/status-panel/` dir, as it's only be used there.
- I think using flex for banner layout would also be a bit cleaner option.
- Panel should have smaller hight. `y` padding seems to be too big. Panel at figma is 90px height, here we have 104px.

paszkowskiDamian: 1. 👍 
2. I'm aware that it's not used in other places but in general it doesn't have anything in common with `market-details` thats why I thought it was a good idea to put it there. Sure I'll move it to `market-details` then. 
3. 👍 
4. Ok I've played a bit with styles and now it's 88px 
image

packages/app/src/features/market-details/components/spark-info-panel/SparkInfoPanel.tsx
```diff
@@ -0,0 +1,32 @@
...
+export function SparkInfoPanel({ title, content }: SparkInfoProps) {
+  return (
+    <div
+      className={cn(
+        'mt-6 py-4 rounded-lg',
+        'col-start-1 col-end-[-1] flex items-center',
+        'bg-spark/10 border-none',
+      )}
+    >
+      <div className={cn('mx-3')}>
```
oskarvu, line 15: By the way, you don't need to use `cn()` for simple class names like in this file. So for example here you can just write: `<div className="mx-3">`. Applicable basically for every `className` in this file. Prettier plugin will handle class name order.

paszkowskiDamian: I thought it was for some type checking or so 😅

packages/app/src/features/market-details/components/spark-info-panel/SparkInfoPanel.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <img src={assets.sparkIcon} alt="Spark logo" style={{ height: '2.75rem' }} />
```
oskarvu, line 16: We also avoid using `style` and use `className` everywhere. If you need custom height, you can do it like this: https://tailwindcss.com/docs/height#arbitrary-values

PR #52, opened 2024-05-17
Enable action settings on production

yivlad:
- Removed `VITE_DEV_ACTIONS_SETTINGS` flag. Now action settings are displayed in all environments.
- Introduced `VITE_FEATURE_CONFIGURABLE_SLIPPAGE` flag - if it's set to `1`, then actions settings will display controls for configuring slippage. This is enabled in dev, stroybook and playwright mode, and disabled on staging and production.
- Removed `Prefer permit` switch from bottom left corner of actions container.
- Fixed application crash when custom slippage input is cleared.
- Removed gas price from bottom right corner of actions container.
- Added E2E test for custom slippage validation:
  - Value is 50%
  - Value is 0%
  - Value is empty


oskarvu (approved):

krzkaczor (approved):
i know it wasn't introduced in this PR but can we tweak description?

"Permits are a way to save gas by allowing a contract to execute multiple actions in a single transaction."

to:

"Permits are a way to save gas by avoiding on-chain approve transactions. Instead signed permits are bundled with another transactions such as deposit or borrow."

krzkaczor (approved):

PR #55, opened 2024-05-27
Migrate prettier to biome

oskarvu:
Migrates prettier to biome formatter.

Notes:
- Biome formatter doesn't support sorting tailwind classes. Biome linter is responsible for that (experimental feature). I added related rule to the linter `useSortedClasses`, but Biome linter is turned off until we migrate from eslint.
- Biome import organizing is turned off until we move away from eslint.


yivlad (approved):

krzkaczor:
FYI: it might be required to tweak your vsc (workspace) settings:
```
"[typescript]": {
    "editor.defaultFormatter": "biomejs.biome"
  },
```

PR #56, opened 2024-05-27
Enable retries in react-query

yivlad:
- Sets number of retries for queries in `react-query` to 3
- Sets the base delay between retry attempts to 150ms.
- Removes duplicate `createSandboxConnector` function.


packages/app/src/test/integration/mockTransport/index.ts
```diff
@@ -31,6 +33,9 @@ export function makeMockTransport(matchers: RpcHandler[]): CustomTransport {
         throw new Error('RPC request not handled: ' + method)
       },
     },
-    { retryCount: 0 },
+    {
+      ...defaultTransportConfig,
```
krzkaczor, line 37: We don't want to use default transport config here. This is totally fake transport.

packages/app/src/config/wagmi/defaults.ts
```diff
@@ -0,0 +1,5 @@
+import { TransportConfig } from 'viem'
+
+export const defaultTransportConfig: Pick<TransportConfig, 'retryCount' | 'retryDelay' | 'timeout'> = {
+  retryCount: 0,
```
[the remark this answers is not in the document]
yivlad: Yes, this is done to avoid retries both in react-query and viem. Leaving both of them on could lead to very long waiting time before query error out.

packages/app/src/config/query-client.ts
```diff
@@ -7,7 +7,8 @@ export const queryClient = new QueryClient({
     queries: {
       gcTime: 1_000 * 60 * 60 * 24, // 24 hours
       refetchOnWindowFocus: true,
-      retry: 0,
+      retry: 3,
+      retryDelay: (attemptIndex) => Math.min(150 * 2 ** attemptIndex, 10_000), // reduce the base delay to 150ms
```
krzkaczor, line 11: Why Math.min? it will be always below 10s with 3 retries.

oskarvu (approved):

packages/app/src/config/query-client.ts
yivlad: Just in case someone overrides the retry prop

krzkaczor (approved):

PR #57, opened 2024-05-28
Improve sentry integration

oskarvu:
- Reports unhandled errors using error boundaries.
- Reports handled errors from queries and mutations.
- Filters out some errors reported to sentry.


packages/app/src/pages/Root.tsx
```diff
@@ -1,19 +1,23 @@
...
+import { ErrorContainer } from '@/features/errors/ErrorContainer'
 import { AppLayout } from '@/ui/layouts/AppLayout'
 import { FallbackLayout } from '@/ui/layouts/FallbackLayout'
 
 export function RootRoute() {
   return (
     <Suspense fallback={<FallbackLayout />}>
       <AppLayout>
-        <Outlet />
+        <ErrorBoundary FallbackComponent={ErrorContainer}>
```
yivlad, line 15: Have you tried adding `errorElement` alongside `RootRoute` in router (https://github.com/marsfoundation/spark-app/blob/cd85cbec3403a9ef7a67abfb25766c4afab6014b/packages/app/src/RootRouter.tsx#L16)? I think this would be preferable solution

oskarvu: I still wanted to display navbar in case of an error. Both for dialogs errors and page errors. Using `errorElement` at `rootRouter` children didn't catch dialog errors (and reporting them to sentry). When you try to add `errorElement` at root path, you won't have navbar then.

I think using single `ErrorBoundary`, that will handle both route and dialog errors makes sense here.

packages/app/src/utils/sentry.ts
```diff
@@ -13,6 +13,12 @@ Sentry.init({
   replaysSessionSampleRate: 0,
   replaysOnErrorSampleRate: 0,
   tracePropagationTargets: [],
+  ignoreErrors: [
+    'User rejected methods',
+    'User disapproved requested methods',
+    'WebSocket request failed',
```
krzkaczor, line 19: we def don't want to ignore errors releated to websockets for now as they are crashing the app.

packages/app/src/utils/sentry.ts
```diff
@@ -13,6 +13,12 @@ Sentry.init({
   replaysSessionSampleRate: 0,
   replaysOnErrorSampleRate: 0,
   tracePropagationTargets: [],
+  ignoreErrors: [
+    'User rejected methods',
+    'User disapproved requested methods',
```
krzkaczor, line 18: do you have idea when this appears?

yivlad (approved):

packages/app/package.json
```diff
@@ -74,6 +74,7 @@
     "react": "^18.2.0",
     "react-confetti": "^6.1.0",
     "react-dom": "^18.2.0",
+    "react-error-boundary": "^4.0.13",
```
krzkaczor, line 77: This component is very simple, i would double think about writing our custom implementation (we don't use 99% of its features anyway...)

packages/app/src/utils/sentry.ts
oskarvu: This will happen, when you reject transaction using mobile wallet (connected by wallet connect).

krzkaczor: okay can you leave a comment next it to explain exactly that?

yivlad (approved):

krzkaczor (approved):

PR #59, opened 2024-05-28
Do not display airdrop amounts

oskarvu:
- Hides airdrop numbers at market table badges and at market details panels
- Fixes airdrop badge value displaying amount only for season 1
- Adds biome rule for unused imports

Note: At `WETH` market details, there is info about spark airdrop. I removed `WETH -> ETH` mapping there, so it's consistent for a page view. If we want to display market details as `ETH` instead of `WETH`, we should do the transformation for the whole view.


packages/app/src/features/market-details/logic/useMarketDetails.ts
```diff
@@ -41,7 +41,7 @@ export function useMarketDetails(): UseMarketDetailsResult {
 
   const isDaiOverview = reserve.token.symbol === marketInfo.DAI.symbol && D3MInfo
 
-  const airdropTokenSymbol = getNativeTokenSymbolIfWrapped(marketInfo.chainId, reserve.token.symbol)
+  const airdropEligibleToken = getNativeTokenSymbolIfWrapped(marketInfo.chainId, reserve.token.symbol)
```
krzkaczor, line 44: why hardcode this when it's already part of the chain config -> airdrop.

krzkaczor (commented):
I am not sure why this PR doesn't modify `packages/app/src/config/chain/index.ts` file with airdrop info.

packages/app/src/config/chain/index.ts
```diff
@@ -52,13 +50,17 @@ const chainConfig: ChainConfig = {
...
-          deposit: NormalizedUnitNumber(6_000_000),
-        },
-        [TokenSymbol('DAI')]: {
-          borrow: NormalizedUnitNumber(24_000_000),
-        },
+      [TokenSymbol('ETH')]: {
+        deposit: [TokenSymbol('SPK')],
+        borrow: [],
+      },
+      [TokenSymbol('WETH')]: {
```
yivlad, line 57: I think we should just do a check if the current asset is wrapped native asset and check for native asset in config

oskarvu: I was thinking about it. But at the same time, this is basically true, cause you will get airdrop for both situations: if you deposit `ETH` and if you deposit `WETH`. Then, logic is simpler down below, cause I don't need to check if asset is wrapped or not.

krzkaczor: Yeah okay we can have it like this for now. We have to fix it once we solve duplicated markets issue...

krzkaczor (approved):

yivlad (approved):

PR #65, opened 2024-05-31
Improve chromatic stories snapshots

oskarvu:
Adds small delay for all the stories to prevent chromatic from taking snapshot in a wrong moment. We are using `useBreakpoint` hook for rendering logic in some places, and without delay, snapshots were taken too early causing nondeterministic behaviour.

It also sets chromatic option for pausing css animations at first frame to make animation snapshots more predictable.


krzkaczor (approved):
@oskarvu to drive this task home, if you notice some undeterministic stories detected on PRs in the upcoming days, please try fix them or just disable given story. 

I feel like having no trust in visual regression setup is very painful for the whole team.

yivlad (approved):

PR #70, opened 2024-06-03
Use getLogs method instead of watching events in useWaitForGnosisTransaction

yivlad:
- Replaces `webSocket` transport with `http` transport in default wagmi config.
- Refactors `useWaitForTransactionReceiptGnosisSafe` to actively query for logs on every new block, instead of using subscription mechanisms.
- Adds `getLogsCall` handler for integration tests.
- Adds `createBlockNumberCallHandler` and `createGetLogsHandler` which effectively allows to simulate blockchain state change in integration tests.

### DIFF
--- a/packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
+++ b/packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
@@ -1,15 +1,17 @@
 import { useState } from 'react'
 import invariant from 'tiny-invariant'
-import { parseAbi } from 'viem'
+import { parseAbiItem } from 'viem'
+import { getLogs } from 'viem/actions'
 import {
   UseWaitForTransactionReceiptParameters,
   UseWaitForTransactionReceiptReturnType,
   useAccount,
+  useClient,
   useWaitForTransactionReceipt,
-  useWatchContractEvent,
+  useWatchBlockNumber,
 } from 'wagmi'
 
-const gnosisAbi = parseAbi(['event ExecutionSuccess(bytes32 txHash, uint256 payment)'])
+const executionSuccessEvent = parseAbiItem('event ExecutionSuccess(bytes32 txHash, uint256 payment)')
 
 /**
  * Like `useWaitForTransactionReceipt`, but works with Gnosis Safe "subtransaction" hashes.
@@ -24,16 +26,22 @@ export function useWaitForTransactionReceiptGnosisSafe(
   const enabled = args.query?.enabled ?? true
   const subTxHash = args.hash
 
+  const client = useClient()
   const [gnosisTxHash, setGnosisTxHash] = useState<`0x${string}` | undefined>()
-  useWatchContractEvent({
-    abi: gnosisAbi,
-    address,
-    eventName: 'ExecutionSuccess',
-    enabled: enabled && !!address && !!subTxHash && gnosisTxHash === undefined,
-    onLogs: (logs) => {
-      if (gnosisTxHash) {
+  useWatchBlockNumber({
+    emitOnBegin: true,
+    enabled,
+    async onBlockNumber(blockNumber) {
+      if (!client) {
         return
       }
+
+      const logs = await getLogs(client, {
+        address,
+        event: executionSuccessEvent,
+        fromBlock: blockNumber - 10n,
+        toBlock: blockNumber,
+      })
 
       for (const log of logs) {
         if (log.args.txHash === subTxHash) {
--- a/packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
+++ b/packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
@@ -1,6 +1,6 @@
 import { skipToken, useQuery } from '@tanstack/react-query'
 import invariant from 'tiny-invariant'
-import { parseAbiItem } from 'viem'
+import { Address, Client, Hash, parseAbiItem } from 'viem'
 import { getLogs, getTransactionReceipt, watchBlockNumber } from 'viem/actions'
 import {
   UseWaitForTransactionReceiptParameters,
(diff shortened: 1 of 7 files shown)


packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
```diff
@@ -1,58 +1,84 @@
...
+      ...waitForTransactionReceiptQueryKey({
+        ...args,
+        chainId: args.chainId ?? chainId,
+      }),
+    ],
+    queryFn:
+      !subTxHash || !client
+        ? skipToken
+        : async ({ signal }) => {
+            const gnosisTxHash = await new Promise<`0x${string}`>((resolve, reject) => {
```
oskarvu, line 42: I think we can use `Hash` from viem instead of `0x${string}` here.

oskarvu (commented):
Did you test it manually also from within safe interface? Or only using wallet connect?

packages/app/src/test/integration/wagmi-config.ts
```diff
@@ -28,6 +28,7 @@ export function createWagmiTestConfig(options: CreateWagmiTestConfigOptions) {
     batch: {
       multicall: false,
     },
+    pollingInterval: 100,
```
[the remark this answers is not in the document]
krzkaczor: Yeah, do we need to change it?

packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
```diff
@@ -1,58 +1,84 @@
...
+): UseWaitForTransactionReceiptReturnType {
   const { address } = useAccount()
-  const enabled = args.query?.enabled ?? true
+  const client = useClient()
+  const chainId = useChainId()
+
   const subTxHash = args.hash
+  const enabled = Boolean(subTxHash && (args.query?.enabled ?? true))
+
+  return useQuery({
```
krzkaczor, line 30: please consider extracting helper functions, it's all a tad too nested. 1 function can be for queryKey, another one for actual logic a query (starting from `async ({ signal }) => {` maybe?)

packages/app/src/domain/hooks/useWaitForTransactionReceiptUniversal.ts
```diff
@@ -38,5 +36,5 @@ export function useWaitForTransactionReceiptUniversal(
     return gnosisTxReceipt
   }
 
-  return { ...receipt, txHash: args.hash }
+  return receipt
```
krzkaczor, line 39: This should be exactly the same thing, right?

packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
```diff
@@ -1,58 +1,84 @@
...
+    queryKey: [
+      'gnosis-safe',
+      ...waitForTransactionReceiptQueryKey({
+        ...args,
+        chainId: args.chainId ?? chainId,
+      }),
+    ],
+    queryFn:
```
krzkaczor, line 38: this is potentially a long-running query. Shouldn't we disable timeouts?

packages/app/src/test/integration/mockTransport/handlers.ts
```diff
@@ -106,6 +113,56 @@ function contractCallError<
...
+  args,
+  blockNumber,
+  transactionHash,
+}: GetLogsCallOptions<TAbiEvent>): RpcHandler {
+  return (method, [params]) => {
+    if (method !== 'eth_getLogs') {
+      return undefined
+    }
+
+    if (!isAddressEqual(params.address, address)) {
```
krzkaczor, line 135: i think it should return undefined then so it's possible to define mutliple `getLogsCall`s

packages/app/src/test/integration/mockTransport/handlers.ts
```diff
@@ -106,6 +113,56 @@ function contractCallError<
...
+    return [
+      {
+        address,
+        topics,
+        data,
+        blockNumber: toHex(blockNumber),
+        transactionHash,
+        // the rest of the parameters are not important for us
+        transactionIndex: '0x41',
+        blockHash: '0x8e8c3f7f1b1d8d6b5b1f6e7a8c9c7d8e7f1b8e7c1b8f7a8e7f1b8e7a8f1b8e7c',
```
krzkaczor, line 158: Instead of hardcoding how about using `getEmptyTxReceipt`'s block data? it's more consistent with other handlers.

packages/app/src/test/integration/wagmi-config.ts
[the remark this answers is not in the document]
yivlad: Why? This is testing config, we want frequently poll for new blocks in tests

packages/app/src/domain/hooks/useWaitForTransactionReceiptUniversal.ts
yivlad: Not exactly, `transactionHash` is accessible in `data` property of the `receipt` object. But generally yrs, it conains the same info so returning `txHash` additionally made no sense

packages/app/src/domain/hooks/useWaitForTransactionReceiptGnosisSafe.ts
yivlad: `react-query` doesn't have concept of  timeouts, the query can be as long running as neccessary

yivlad:
> Did you test it manually also from within safe interface? Or only using wallet connect?

Tested both in the Safe App and using Wallet Connect

packages/app/src/test/integration/wagmi-config.ts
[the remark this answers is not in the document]
krzkaczor: aaah right, i didn't realize it's a test config 👍 it's fine then

packages/app/src/domain/hooks/useWaitForTransactionReceiptUniversal.ts
krzkaczor: hmm yes, i think we never used that, can you double check?

krzkaczor (approved):

oskarvu (approved):

packages/app/src/domain/hooks/useWaitForTransactionReceiptUniversal.ts
yivlad: Yes, it was never used

PR #74, opened 2024-06-05
Add E2E test for switching assets in deposit to savings dialog

yivlad:


packages/app/src/features/dialogs/savings/deposit/SavingsDepositDialog.test-e2e.ts
```diff
@@ -445,4 +474,6 @@ test.describe('Savings deposit dialog', () => {
       await actionsContainer.expectNextActionEnabled()
     })
   })
+
+  test.describe('', () => {})
```
oskarvu, line 478: Can be removed.

oskarvu (approved):

packages/app/src/features/dialogs/savings/deposit/SavingsDepositDialog.test-e2e.ts
```diff
@@ -84,6 +84,35 @@ test.describe('Savings deposit dialog', () => {
 
       await savingsPage.expectCurrentWorth('99.00')
     })
+
+    test('can switch from USDC to xDAI', async ({ page }) => {
```
krzkaczor, line 88: imo this should be called "can switch between tokens" and should go from dai -> usdc ->usdt -> dai (soon each token will have slightly different action plan).

krzkaczor (approved):

PR #82, opened 2024-06-07
Add transaction details for native Maker routes on Savings page

yivlad:
image

Works also for withdraw


packages/app/src/features/dialogs/savings/common/components/MakerTransactionOverview.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </div>
+      {!isLast && <img src={assets.arrowRight} className="h-3.5 w-3.5" />}
+      <div className="justify-self-end text-basics-dark-grey text-sm">{item.token.formatUSD(item.value)}</div>
+    </div>
+  )
+}
+
+function MakerBadge({ token }: { token: Token }) {
+  return (
+    <div className="flex flex-row items-center gap-1.5 rounded-lg bg-emerald-300/10 px-2.5 py-1.5 text-emerald-400 text-sm">
```
oskarvu, line 90: Can you add custom green color? One from tailiwind is quite different than what we have in figma.

packages/app/src/features/dialogs/savings/deposit/views/SavingsDepositView.stories.tsx
oskarvu, line 1: Can we add story for native route here as well?

packages/app/src/features/dialogs/savings/deposit/logic/createTxOverview.ts
```diff
@@ -82,3 +71,38 @@ export function createTxOverview({
...
+  const route: RouteItem[] = [
+    {
+      token: marketInfo.DAI,
+      value: daiValue,
+      usdValue: daiValue,
+    },
+    {
+      token: marketInfo.sDAI,
+      value: sDAIValue,
+      usdValue: savingsInfo.convertSharesToDai({ shares: sDAIValue }),
```
krzkaczor, line 97: here let's just us usdValues to get rid of rounding issues

oskarvu (approved):

krzkaczor:
Final comments: 
1) Dialog height shouldn't change after inputing non-zero value: 
image
image

2) [This story](https://www.chromatic.com/test?appId=6629151a3bd0c7efaaab1e03&id=6666e8af53d33e64d82d6bf4) should use DAI to be correct.

krzkaczor (approved):

PR #86, opened 2024-06-11
Improve scripts and useTimestamp tests

krzkaczor:
* run `pnpm run fix` to run all fixes before commiting new change
* CI (test) faster 20s

Closes FRO-743


oskarvu (approved):

packages/app/src/utils/useTimestamp.test.ts
```diff
@@ -40,15 +41,17 @@ describe('useTimestamp', () => {
...
-    await waitFor(() => {
-      expect(result.current.timestampInMs).toBeGreaterThan(initialTimestampInMs)
-      expect(result.current.timestamp).toBeGreaterThan(initialTimestamp)
-    })
+    await vi.advanceTimersByTimeAsync(2_000)
+
+    expect(result.current.timestampInMs).toBeGreaterThan(initialTimestampInMs)
+    expect(result.current.timestamp).toBeGreaterThan(initialTimestamp)
+
+    vi.restoreAllMocks()
```
yivlad, line 55: Put it in `afterAll/afterEach`? This line won't run if the assertions before fail

krzkaczor: hmm good point... i didn't want to mock all tests because mocking timers is not compatible with `waitFor`. but afterEach will work...

yivlad (approved):

PR #88, opened 2024-06-11
Do not reload sandbox chain on refresh

yivlad:


oskarvu (approved):

packages/app/.env.playwright
```diff
@@ -6,4 +6,5 @@ VITE_FEATURE_CONFIGURABLE_SLIPPAGE=1
 # development
 VITE_DEV_NATIVE_ROUTES=1
 
+VITE_WALLET_CONNECT_ID='bd1843f2419f5d4c758366c55f9a556c'
```
krzkaczor, line 9: we need that?

packages/app/src/config/wagmi/config.default.ts
```diff
@@ -23,6 +23,40 @@ export function getConfig(sandboxNetwork?: SandboxNetwork): Config {
     [gnosis.id]: http('https://rpc.ankr.com/gnosis'),
   }
 
+  const defaultStorage = createStorage({
+    storage: typeof window !== 'undefined' && window.localStorage ? window.localStorage : noopStorage,
+  })
+  const storage: typeof defaultStorage = {
+    ...defaultStorage,
+    getItem: async (key) => {
```
krzkaczor, line 31: please extract to a function with meaninful name or add a comment explaining what it does and we need it.

packages/app/.env.playwright
yivlad: Yes, we load default (not E2E) config in the added test

packages/app/src/config/wagmi/storage.ts
```diff
@@ -0,0 +1,43 @@
...
+  })
+  const storage: typeof defaultStorage = {
+    ...defaultStorage,
+    getItem: wrapGetItem(defaultStorage),
+  }
+
+  return storage
+}
+
+function wrapGetItem(defaultStorage: ReturnType<typeof createStorage>): any {
```
krzkaczor, line 17: Please leave a comment why we do this like: 
> to avoid autoreconnection to sandbox network, we explictly remove it from saved connections

krzkaczor (approved):

PR #91, opened 2024-06-12
Add DAI native withdrawal e2e tests

oskarvu:
Closes FRO-719


packages/app/src/features/dialogs/savings/common/e2e/SavingsDialog.PageObject.ts
```diff
@@ -1,10 +1,13 @@
+import { ActionsPageObject } from '@/features/actions/ActionsContainer.PageObject'
 import { testIds } from '@/ui/utils/testIds'
 import { Page, expect } from '@playwright/test'
-import { DialogPageObject } from '../../common/Dialog.PageObject'
+import { DialogPageObject } from '../../../common/Dialog.PageObject'
 
-export class SavingsDepositDialogPageObject extends DialogPageObject {
-  constructor(page: Page) {
-    super(page, /Deposit to Savings/)
+export class SavingsDialogPageObject extends DialogPageObject {
```
yivlad, line 6: I would tbh keep two different classes `SavingsDepositDialogPageObject` and `SavingsWithdrawDialogPageObject` , just extract common functionality into base class `SavingsDialogPageObject`

krzkaczor: But then we would need to introduce inheritance. I think i like it as it is (at least for now) both dialogs are extremely similiar.

krzkaczor (approved):

yivlad (approved):

PR #95, opened 2024-06-13
Enable native DAI routes for production

krzkaczor:


packages/app/.env.production
```diff
@@ -8,7 +8,7 @@ VITE_FEATURE_TOP_BANNER=1
 VITE_FEATURE_CONFIGURABLE_SLIPPAGE=0
 
 # development
-VITE_DEV_NATIVE_ROUTES=0
+VITE_DEV_NATIVE_ROUTES=1
```
oskarvu, line 11: Maybe we should remove flag completely since it's not in development anymore?

krzkaczor: Let's do it few days after deployment.

oskarvu (approved):

yivlad (approved):

PR #105, opened 2024-06-17
Add native xDai deposit and use separate actions for other native deposits

oskarvu:
Closes FRO-736


packages/app/src/features/actions/flavours/native-dai-deposit/NativeDaiDepositActionRow.tsx
yivlad, line 1: I would group added actions into one folde (`native-sdai-deposit`), and then introduce `common` folder and reuse ui. I imagine that `NativeUSDCDepositActionRow.tsx` should only re-export the component from `common` folder. Also I think naming is a bit misleading - I would name each flavour `native-dai-to-sdai-deposit`, or even just `dai-to-sdai-deposit`.

oskarvu: Agree on naming. Not sure about putting staff in some grouping folder. Kinda like this linear structure, even if some ui is duplicated. @krzkaczor What do you think?

oskarvu: Renamed actions and decided to reuse ui for rows.

yivlad (approved):

krzkaczor (approved):

PR #109, opened 2024-06-17
Add tests for AssetInput component

yivlad:


packages/app/src/ui/organisms/multi-selector/MultiSelector.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+        maxSelectedFieldName="isMaxSelected"
+        max={max}
+      />
+    </Form>
+  )
+}
+
+describe(ControlledMultiSelectorAssetInput.name, () => {
+  test('Changes value in input', async () => {
+    const value = NormalizedUnitNumber(12345)
```
oskarvu, line 59: Maybe just `const value = '12345'`?

oskarvu (approved):

packages/app/src/ui/organisms/multi-selector/MultiSelector.test.tsx
yivlad: I think it's better to be consistent and have `NormalizedUnitNumber` everywhere

packages/app/src/ui/organisms/multi-selector/MultiSelector.tsx
```diff
@@ -127,6 +127,21 @@ export function ControlledMultiSelectorAssetInput({
             setMax={setMaxValue ?? setMaxSelectedField}
             isMaxSelected={maxSelectedFieldName && isMaxSelected}
             onChange={(e) => {
+              if (e.target.value === MAX_VALUE_PLACEHOLDER.slice(0, -1)) {
+                // backspace was pressed while MAX was selected
+                ;(setMaxValue ?? setMaxSelectedField)?.()
+                return
+              }
+
+              e.target.value = e.target.value.replace(/,/g, '.')
```
krzkaczor, line 136: I am not sure if it's good idea to extract this logic from AssetInput. Don't you think we might want to use it directly in the future?

yivlad: Good point

krzkaczor (approved):

PR #110, opened 2024-06-18
Fix debt ceiling unit for isolated assets

yivlad:


oskarvu (approved):

oskarvu:
I would just change debt ceiling values at stories for more realistic ones.

krzkaczor (approved):

PR #112, opened 2024-06-18
Add custom lint rule

krzkaczor:
Closes FRO-766


packages/app/scripts/verify-svgs.ts
```diff
@@ -0,0 +1,34 @@
...
+const assetsPath = path.join(__dirname, '../src/ui/assets')
+
+const paths = await globby(['**/*.svg'], { cwd: assetsPath, absolute: true })
+const erroredPaths: string[] = []
+
+// biome-ignore lint/suspicious/noConsoleLog: <explanation>
+console.log(`Verifying ${paths.length} SVGs...`)
+for (const p of paths) {
+  const content = await readFileSync(p, 'utf8')
+  if (content.includes('<?xml')) {
```
yivlad, line 22: What about using something like [this](https://github.com/sindresorhus/is-svg/blob/main/index.js) and ensuring that parsed xml object has only one key? And the key's name is `svg`

packages/app/scripts/verify-svgs.ts
```diff
@@ -0,0 +1,37 @@
...
+const errors: {msg: string, file: string} [] = []
+
+// biome-ignore lint/suspicious/noConsoleLog: <explanation>
+console.log(`Verifying ${paths.length} SVGs...`)
+for (const p of paths) {
+  const content = await readFileSync(p, 'utf8')
+  if (content.includes('<?xml')) {
+    errors.push({file: p, msg: 'Found <?xml tag'})
+  }
+  if (content.startsWith('\n')) {
```
yivlad, line 25: Maybe whitespace in general?

yivlad (approved):

oskarvu (approved):

PR #118, opened 2024-06-19
Fix isolated asset total debt display

yivlad:


packages/app/src/features/market-details/logic/makeMarketOverview.ts
```diff
@@ -16,6 +17,10 @@ export function makeMarketOverview({ reserve, marketInfo }: MakeMarketOverviewPa
     marketInfo,
     token: reserve.token.symbol,
   })
+  const debt =
```
krzkaczor, line 20: this is used only in isolation mode, i suggest:
* rename it to isolationModeCurrentDebt (and isolationModeDebtCeiling)
* make it undefined when not in isolation mode
(could also be a union type but simple fix will work for now)

krzkaczor (approved):

oskarvu (approved):

PR #123, opened 2024-06-20
Add future market info calculations

yivlad:


packages/app/src/features/dialogs/repay/logic/useRepayDialog.ts
```diff
@@ -39,9 +39,13 @@ export interface UseRepayDialogResult {
 export function useRepayDialog({ initialToken }: UseRepayDialogOptions): UseRepayDialogResult {
   const { aaveData } = useAaveDataLayer()
   const { marketInfo } = useMarketInfo()
-  const nativeAssetInfo = getNativeAssetInfo(marketInfo.chainId)
+  const { marketInfo: marketInfoIn1Epoch } = useMarketInfo({ timeAdvance: 30 * 60 })
```
oskarvu, line 42: I wonder if we could use it like this:
```const [marketInfo, marketInfoIn1Epoch] = useMarketInfo({timeAdvance: ['now', '1epoch']})```
Just an idea, cause calling hooks two times is ok.

File: packages/app/src/domain/market-info/aave-data-layer/query.ts [hunk 2]
```diff
@@ -11,33 +11,69 @@ import {
...
+    queryKey: aaveDataLayerQueryKey({ chainId, account }),
+    queryFn: aaveDataLayerQueryFn({
+      uiPoolDataProvider,
+      lendingPoolAddressProvider,
+      uiIncentiveDataProvider,
+      wagmiConfig,
+      chainId,
+      account,
+    }),
+    select: aaveDataLayerSelectFnWithCache({ timeAdvance }),
```
oskarvu, line 55: I'm curious if would't be enough to just use React.useCalback to memoise selector function?

packages/app/src/domain/market-info/useMarketInfo.ts
```diff
@@ -19,15 +19,14 @@ export function useMarketInfo(params: UseMarketInfoParams = {}): UseMarketInfoRe
...
-  const nativeAssetInfo = getNativeAssetInfo(chainId)
 
   const res = useSuspenseQuery({
     ...aaveDataLayer({
       wagmiConfig,
       chainId,
       account: address,
     }),
-    select: (aaveData) => marketInfo(aaveData, nativeAssetInfo, chainId),
+    select: marketInfoSelectFn({ timeAdvance: params.timeAdvance }),
```
oskarvu, line 29: Similarly here, maybe using `React.useCallback` would be enough.

packages/app/src/features/dialogs/repay/logic/useRepayDialog.ts
oskarvu: Also, it would be better to pass timeAdvance option not as a number, but as some predefined values/variants.

packages/app/src/domain/market-info/aave-data-layer/query.ts
```diff
@@ -62,9 +98,44 @@ export function aaveDataLayer({ wagmiConfig, chainId, account }: AaveDataLayerAr
...
-    const [[reserves, baseCurrencyInfo], reservesIncentiveData, [userReserves, userEmodeCategoryId]] = data
+    return {
+      contractData,
+      chainId,
+      lendingPoolAddressProvider,
+    }
+  }
+}
+
+const aaveDataLayerSelectFnCache = new Map<number, ReturnType<typeof aaveDataLayerSelectFn>>()
```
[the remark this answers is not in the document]
yivlad: WeakMap requires keys to be referential object (function or object),so it's not applicable here

packages/app/src/domain/market-info/aave-data-layer/query.ts
oskarvu, line 55: I was thinking again about this cache. We could remove it, have flexibility of custom `timeAdvance` parameter, and no re-renders.

Here, you would export normal selector, without cache. And in places, where `aaveDataLayer` options are used, you would need to wrap this select with `React.useCallback`. Similarly with `marketInfo` selector.

packages/app/src/domain/market-info/useMarketInfo.ts
oskarvu, line 29: Here you would just need to wrap normal (without cache) market info select fn with `useCallback`.

packages/app/src/features/navbar/logic/useTotalBalance.ts
```diff
@@ -30,7 +28,7 @@ export function useTotalBalance(): BalanceInfo {
       account: address && CheckedAddress(address),
       chainId,
     }),
-    select: (aaveData) => marketInfoSelect(aaveData, nativeAssetInfo, chainId),
+    select: marketInfoSelectFn(),
```
oskarvu, line 31: similarly here, just wraping with `useCallback` non cashed selector.

krzkaczor (approved):

PR #126, opened 2024-06-21
Add E2E test for time-based max repay value calculations

yivlad:


packages/app/src/features/dialogs/repay/RepayDialog.test-e2e.ts
```diff
@@ -250,12 +261,112 @@ test.describe('Repay dialog', () => {
...
+        BaseUnitNumber(
+          NormalizedUnitNumber(daiToBorrow).times(daiDebtIncreaseIn1Epoch).shiftedBy(DAI_DECIMALS).minus(1),
+        ),
+      )
+
+      const dashboardPage = new DashboardPageObject(page)
+
+      // The code below is needed to refresh the balances.
+      // Doing reload breaks the current debt amount for some reason.
+      // Any other solution tried had the same issue.
```
krzkaczor, line 289: did you try reloading and then injecting things that we inject in `setup`?

oskarvu: Maybe having separate setup for each test wouldn't be that bad.

packages/app/src/features/dialogs/repay/RepayDialog.test-e2e.ts
```diff
@@ -212,19 +215,27 @@ test.describe('Repay dialog', () => {
   })
 
   test.describe('Position when borrowed asset was not in user wallet before', () => {
+    const DAI_ADDRESS = '0x6B175474E89094C44Da98b954EedeAC495271d0F'
```
krzkaczor, line 218: The whole describe block should be extracted from the parent block because the parent block already contains `beforeEach` which is totally ignored right now and only adds confusion.

yivlad: I think the `describe` block here is the second nested one, and it has only one before each.

packages/app/src/features/dialogs/repay/RepayDialog.test-e2e.ts
```diff
@@ -212,50 +216,137 @@ test.describe('Repay dialog', () => {
...
       await actionsContainer.acceptAllActionsAction(2)
       await repayDialog.expectSuccessPage([repay], fork)
 
-      await screenshot(repayDialog.getDialog(), 'repay-dialog-whole-balance-dai-success')
+      await repayDialog.viewInDashboardAction()
+
+      await dashboardPage.expectNonZeroAmountInBorrowTable(repay.asset)
+    })
+
+    test('can repay if balance gt debt but lt debt after 1 epoch', async ({ page }) => {
```
krzkaczor, line 280: why this test doesnt repay the whole debt? as it's stated in the test name balance > debt. 

Dai debt is `daiToBorrow`, and dai balance is daiToBorrowInOneEpoch (15 min?) - 1 wei. This should be enough to pay whole debt, no?

oskarvu (approved):

krzkaczor (approved):

PR #128, opened 2024-06-24
Savings validation e2e tests

oskarvu:
Closes FRO-723

### DIFF
--- a/packages/app/src/features/dialogs/savings/deposit/e2e/mainnet/DepositDAI.test-e2e.ts
+++ b/packages/app/src/features/dialogs/savings/deposit/e2e/mainnet/DepositDAI.test-e2e.ts
@@ -77,7 +77,60 @@ test.describe('Validation', () => {
   let savingsPage: SavingsPageObject
   let depositDialog: SavingsDialogPageObject
 
-  test.beforeEach(async ({ page }) => {
+  test.describe('Input value exceeds balance', () => {
+    test.beforeEach(async ({ page }) => {
+      await setup(page, fork, {
+        initialPage: 'savings',
+        account: {
+          type: 'connected',
+          assetBalances: {
+            ETH: 1,
+            DAI: 100,
+          },
+        },
+      })
+
+      savingsPage = new SavingsPageObject(page)
+      await savingsPage.clickDepositButtonAction('DAI')
+
+      depositDialog = new SavingsDialogPageObject({ page, type: 'deposit' })
+      await depositDialog.fillAmountAction(10_000)
+    })
+
+    test('displays validation error', async () => {
+      await depositDialog.expectAssetInputError(depositValidationIssueToMessage['exceeds-balance'])
+    })
+
+    test('actions are disabled', async () => {
+      await depositDialog.actionsContainer.expectDisabledActions([
+        { type: 'approve', asset: 'DAI' },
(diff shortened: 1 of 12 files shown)


packages/app/src/features/dialogs/savings/deposit/e2e/mainnet/DepositDAI.test-e2e.ts
```diff
@@ -70,3 +71,64 @@ test.describe('Deposit DAI on Mainnet', () => {
...
+
+    savingsPage = new SavingsPageObject(page)
+    await savingsPage.clickDepositButtonAction('DAI')
+
+    depositDialog = new SavingsDialogPageObject({ page, type: 'deposit' })
+  })
+
+  test('displays validation error', async () => {
+    await depositDialog.fillAmountAction(200)
+    await depositDialog.expectAssetInputError(depositValidationIssueToMessage['exceeds-balance'])
```
yivlad, line 100: I think we should hardcode the error messages here. It's ok if we'll get "notified" when the text changes

packages/app/src/features/dialogs/savings/deposit/e2e/mainnet/DepositDAI.test-e2e.ts
```diff
@@ -70,3 +71,64 @@ test.describe('Deposit DAI on Mainnet', () => {
...
+    await depositDialog.fillAmountAction(0)
+    await depositDialog.expectAssetInputError(depositValidationIssueToMessage['value-not-positive'])
```
yivlad, line 102: Same as above

packages/app/src/features/dialogs/savings/deposit/e2e/mainnet/DepositDAI.test-e2e.ts
oskarvu: Was following convention @krzkaczor introduced after implementing validation.

krzkaczor: IMO this is the way to go.

packages/app/src/features/dialogs/savings/deposit/e2e/mainnet/DepositDAI.test-e2e.ts
```diff
@@ -70,3 +71,64 @@ test.describe('Deposit DAI on Mainnet', () => {
...
+  test('actions are disabled', async () => {
+    await depositDialog.fillAmountAction(200)
+    await depositDialog.actionsContainer.expectDisabledActions([
+      { type: 'approve', asset: 'DAI' },
+      { type: 'daiToSDaiDeposit', asset: 'DAI' },
+    ])
+  })
+
+  test('displays sensible tx overview', async () => {
+    await depositDialog.fillAmountAction(10_000)
```
krzkaczor, line 114: what i don't like about this test is that it doesnt assert that validation failed (it uses a different value).  Also test names are not descriptive enough. I would structurize these tests like this:
* Validation:
  *  When value exceeds balance (here we already input 10_000 in before each):
      * it displays validation error
      * it disables actions
      * it simulates route and apy
  * it displays validation error on zero when dirty

yivlad (approved):

krzkaczor (approved):

PR #129, opened 2024-06-24
Add reading user rewards and useClaimAllRewards hook

yivlad:


packages/app/src/domain/market-info/aave-data-layer/query.ts
```diff
@@ -30,6 +30,7 @@ export interface AaveDataLayerArgs extends AaveDataLayerQueryKeyArgs {
 export type AaveData = ReturnType<ReturnType<typeof aaveDataLayerSelectFn>>
 export type AaveUserSummary = AaveData['userSummary']
 export type AaveUserReserve = AaveUserSummary['userReservesData'][number]['reserve']
+export type AaveUserRewaed = AaveData['userRewards']
```
oskarvu, line 33:
```suggestion
export type AaveUserRewards = AaveData['userRewards']
```

packages/app/src/domain/market-info/marketInfo.ts
```diff
@@ -347,6 +355,22 @@ export function marketInfoSelectFn({ timeAdvance }: MarketInfoSelectFnParams = {
       isolationModeState: determineIsolationModeState(rawAaveData.userSummary, reserves),
       siloBorrowingState: determineSiloBorrowingState(userPositions),
     }
+
+    const userRewards: UserReward[] = Object.entries(rawAaveData.userRewards)
+      .reduce(
+        (acc, [_, value]) => [
+          ...acc,
+          {
+            value: NormalizedUnitNumber(value.claimableRewards.dp(0).shiftedBy(-value.rewardTokenDecimals)),
```
oskarvu, line 364: Why do you need to use `dp(0)`? Are claimable rewards having decimals? I'm curious, cause you're shifting value later, so maybe `claimableRewards` is in base unit number format?

packages/app/src/domain/rewards/useClaimAllRewards.ts
```diff
@@ -0,0 +1,46 @@
...
+  const chainId = useChainId()
+  const { address: account } = useAccount()
+
+  return useWrite(
+    {
+      address: incentiveControllerAddress,
+      abi: incentiveControllerAbi,
+      functionName: 'claimAllRewards',
+      args: [assets, account!],
+      enabled: enabled ?? Boolean(account && assets.length > 0),
```
oskarvu, line 32: Only if `enabled` is undefined, right hand side of the `??` operator will be evaluated. So for example, if you set `enabled` to `true`, but there will be no `account` or `assets`, hook will still be enabled. Was it intended?

packages/app/src/config/abis/incentiveControllerAbi.ts
oskarvu, line 1: Is it added here instead of to wagmi config cause each reward program can have different incentive controller address, but all of them will have same abi?

yivlad: Yes, we don't know in advance the incentive controller address.

oskarvu (approved):

packages/app/src/domain/market-info/marketInfo.ts
```diff
@@ -347,6 +355,22 @@ export function marketInfoSelectFn({ timeAdvance }: MarketInfoSelectFnParams = {
       isolationModeState: determineIsolationModeState(rawAaveData.userSummary, reserves),
       siloBorrowingState: determineSiloBorrowingState(userPositions),
     }
+
+    const userRewards: UserReward[] = Object.entries(rawAaveData.userRewards)
+      .reduce(
```
krzkaczor, line 360: why reduce instead of map?

packages/app/src/domain/market-info/marketInfo.ts
krzkaczor: dp(0) i guess makes sense, but shouldn't you use a token method to get normalized unit number?

krzkaczor (approved):

PR #133, opened 2024-06-26
Add markets page e2e tests

oskarvu:
Closes FRO-774


packages/app/src/pages/Markets.PageObject.ts
```diff
@@ -0,0 +1,154 @@
...
+      await expect(tileLocator).toContainText(tile.value)
+    }
+  }
+  async expectActiveMarketsTable(rows: MarketsTableRow[]): Promise<void> {
+    await this.expectMarketsTable(rows, this.locateActiveMarketsRows())
+  }
+  async expectFrozenMarketsTable(rows: MarketsTableRow[]): Promise<void> {
+    await this.expectMarketsTable(rows, this.locateFrozenMarketsTableBody())
+  }
+  async expectMarketsTable(rows: MarketsTableRow[], rowsLocator: Locator): Promise<void> {
```
yivlad, line 36: I'm concerned that this will not work if we have to wait the table structure to update. Can we use the approach with giving each row a data-testId and then querying for it?

oskarvu: In case of markets page, having table = having table structure defined, so we don't wait for anything to update. I can always add ids to the rows, but I really don't think it's needed here.

oskarvu: I added it anyway ;)

yivlad (approved):

krzkaczor (approved):

PR #140, opened 2024-07-01
Display 404 when going to non-existent market details

yivlad:


packages/app/src/features/errors/ErrorBoundary.tsx
```diff
@@ -29,6 +29,14 @@ export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundarySt
 
   render() {
```
oskarvu, line 30: nit, optional: no nested ifs and grouped falback case.


```suggestion
render() {
  const { hasError } = this.state;
  const { children, fallback } = this.props;

  if (!hasError) {
    return children;
  }

  if (!fallback) {
    throw this.state.error;
  }

  return typeof fallback === 'function'
    ? fallback({ error: this.state.error })
    : fallback;
}
```

packages/app/src/features/errors/RouterErrorFallback.tsx
```diff
@@ -7,12 +7,21 @@ import { NotConnected } from './NotConnected'
...
   const error = useRouteError()
+
+  return <ErrorFallback error={error} fullScreen={fullScreen} />
+}
+
+export interface ErrorFallbackProps {
+  fullScreen?: boolean
+  error?: any
+}
+export function ErrorFallback({ fullScreen, error }: ErrorFallbackProps) {
```
oskarvu, line 24: Maybe worth to put in a separate file if it's reused.

oskarvu (approved):

krzkaczor (approved):

packages/app/src/features/errors/ErrorBoundary.tsx
```diff
@@ -28,10 +28,18 @@ export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundarySt
...
-    if (this.state.hasError) {
-      return this.props.fallback
+    if (!this.state.hasError) {
+      return this.props.children
     }
 
-    return this.props.children
+    if (!this.props.fallback) {
```
krzkaczor, line 35: this case is not possible according to prop types 🤔

yivlad: Right, wondering why linter didn't catch that

PR #142, opened 2024-07-02
Display actual max value instead of placeholder in repay dialog

yivlad:


oskarvu (approved):

packages/app/src/ui/molecules/asset-input/AssetInput.tsx
```diff
@@ -40,7 +40,7 @@ export const AssetInput = forwardRef<HTMLInputElement, AssetInputProps>(
       onChange,
       variant = 'crypto',
       walletIconLabel,
-      isMaxSelected,
```
krzkaczor, line 43: i think this flag might become handy to disable MAX button SOON.

packages/app/src/features/dialogs/repay/logic/extractRepayMaxValueFromForm.ts
```diff
@@ -1,28 +1,25 @@
...
 
-export interface ProcessRepaymentAssetParams {
+export interface ExtractRepayMaxValueFromFormParams {
   formValues: DialogFormNormalizedData
   marketInfo: MarketInfo
   walletInfo: WalletInfo
 }
 
-export function processRepaymentAsset({
+export function extractRepayMaxValueFromForm({
```
krzkaczor, line 13: this name suggests that max value uses form values which is not true. Maybe a better name would be `getRepayMaxValueForFormAsset` (but i am not fully convinced that i like it...)?

packages/app/src/features/dialogs/repay/logic/form.ts
```diff
@@ -76,7 +79,7 @@ export function getFormFieldsForRepayDialog(
       balance: walletInfo.findWalletBalanceForSymbol(symbol),
     },
     changeAsset,
-    maxValue: NormalizedUnitNumber(10),
+    maxValue: NormalizedUnitNumber(repayMaxValue.dp(6, BigNumber.ROUND_DOWN)),
```
krzkaczor, line 82: please remember to extract it to a function like "formatMaxValuePlaceholder" if we end up using it in multiple places.

krzkaczor, line 82: shouldn't it be called `placeholderMaxValue` ie. a value that is not really used but we show it to the user?

packages/app/src/ui/molecules/asset-input/AssetInput.tsx
```diff
@@ -41,11 +43,24 @@ export const AssetInput = forwardRef<HTMLInputElement, AssetInputProps>(
...
       walletIconLabel,
       showMaxPlaceholder,
+      isMaxSelected,
       ...rest
     },
     ref,
   ) => {
     assert(!(balance && !token), 'token should be defined if balance is defined')
+    const inputValue = (() => {
+      if (showMaxPlaceholder && isMaxSelected) {
```
krzkaczor, line 53: Please refactor to avoid needlessly parsing strings to BigNumbers :
```
if (isMaxSelected) {
	if (showMaxPlaceholder)
	//parse and cut number}
}
return value
```

packages/app/src/ui/organisms/multi-selector/MultiSelector.tsx
```diff
@@ -99,6 +99,12 @@ export function ControlledMultiSelectorAssetInput({
             setValue(maxSelectedFieldName, !isMaxSelected, {
               shouldValidate: true,
             })
+            if (isMaxSelected) {
```
krzkaczor, line 102: this is gonna be a dead code (and removed) once we disable MAX buttons after clicking them once.

krzkaczor (approved):

PR #143, opened 2024-07-02
Prepare withdraw dialog to serve as send dialog when needed

oskarvu:
Closes FRO-779

Introduced separate address form to handle receiver. I made that decision to prevent complicating logic behind core withdraw functionality and to treat `send mode` more like a plugin/extension.

Related PR (with validation), that adds validation and gives more context to design choices I made is available here: https://github.com/marsfoundation/spark-app/pull/146


krzkaczor (approved):

packages/app/src/features/dialogs/savings/withdraw/components/form/ControlledAddressInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface ControlledAddressInputProps {
+  form: UseFormReturn<ReceiverFormSchema>
+  blockExplorerUrl: string | undefined
+}
+
+export function ControlledAddressInput({ form, blockExplorerUrl }: ControlledAddressInputProps) {
+  return (
+    <Controller
+      control={form.control}
+      name="receiver"
```
yivlad, line 14: I think this should be passed as a prop to make it consistent with `ControledMultiSelector`

yivlad (approved):

packages/app/src/features/dialogs/savings/withdraw/components/form/ControlledAddressInput.tsx
oskarvu: `name` value is type here, so it doesn't make much sense to pass as additional prop.

PR #145, opened 2024-07-02
Introduce minRemainingNativeAssetBalance and use it when repaying max

yivlad:


packages/app/src/domain/market-info/marketInfo.ts
```diff
@@ -147,7 +147,7 @@ export class MarketInfo {
     public readonly timestamp: number,
     public readonly chainId: number,
     public readonly userRewards: UserReward[],
-    private readonly nativeAssetInfo: NativeAssetInfo,
+    public readonly nativeAssetInfo: NativeAssetInfo,
```
oskarvu, line 150: I'm not sure about this. Why making two sources of the same thing, when we can easily and everywhere read this from config?

packages/app/src/features/dialogs/repay/RepayDialog.test-e2e.ts
```diff
@@ -562,5 +562,33 @@ test.describe('Repay dialog', () => {
       await dashboardPage.expectBorrowedAssetsToBeEmpty()
       await screenshot(page, 'repay-dialog-nothing-to-repay')
     })
+
+    test('when repaying native asset leave retain some in wallet', async ({ page }) => {
```
oskarvu, line 566:
```suggestion
    test('when repaying native asset leave some balance in wallet', async ({ page }) => {
```

oskarvu (approved):

PR #146, opened 2024-07-03
Add receiver form validation

oskarvu:
Closes FRO-785

### DIFF
--- a/packages/app/src/domain/hooks/useIsSmartContract.ts
+++ b/packages/app/src/domain/hooks/useIsSmartContract.ts
@@ -1,15 +1,15 @@
 import { Address } from 'viem'
 import { useBytecode } from 'wagmi'
 
-interface UseIsSmartContractResult {
+export interface UseIsSmartContractResult {
   isSmartContract?: boolean
-  isLoading: boolean
+  isPending: boolean
 }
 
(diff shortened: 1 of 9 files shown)


packages/app/src/domain/hooks/useIsSmartContract.ts
```diff
@@ -0,0 +1,15 @@
+import { Address } from 'viem'
+import { useBytecode } from 'wagmi'
+
+export interface UseIsSmartContractResult {
```
krzkaczor, line 4: why do you need isPending here? why not to use a common pattern of returning T or undefined when still loading? ie. boolean | undefined in this case?

krzkaczor (approved):

packages/app/src/domain/hooks/useIsSmartContract.ts
oskarvu: I need `isPending` to disable actions until this hook has result. This covers error case. `isPending` is `false` in case of error. Without `isPending` and in case of error, hook would return undefined indefinitely.

I also like having concrete indicator if query is in progress rather than inferring it from data.

yivlad (approved):

PR #147, opened 2024-07-03
Disable MAX button after click in AssetInput component

yivlad:


oskarvu:
One comment. When testing max button behaviour I noticed, that it doesn't behave same in all inputs. I think it would be nice to have same behaviour everywhere, but I'm not sure if we should do it in this PR. cc @krzkaczor

krzkaczor (commented):
What about deleteting this? https://github.com/marsfoundation/spark-app/pull/142/files#diff-7fd2ae79d398b704339c9d10f6e9261e03189f99fe7a890c050d7cbacbe3f0b7R102

oskarvu (approved):

krzkaczor (approved):

PR #148, opened 2024-07-03
Add debt and supply change indicators in dialogs

yivlad:


packages/app/src/features/dialogs/common/components/TokenValueChange.tsx
oskarvu, line 1: I wonder if we should use it also in deposit/borrow overviews? It would be good to see of how much debt is increasing when borrowing.

packages/app/src/ui/molecules/asset-input/AssetInput.tsx
```diff
@@ -50,15 +50,13 @@ export const AssetInput = forwardRef<HTMLInputElement, AssetInputProps>(
   ) => {
     assert(!(balance && !token), 'token should be defined if balance is defined')
     const inputValue = (() => {
-      if (isMaxSelected) {
-        if (showMaxPlaceholder) {
-          return MAX_VALUE_PLACEHOLDER
-        }
+      if (isMaxSelected && showMaxPlaceholder) {
```
oskarvu, line 53: Didn't you change this in one of your previous PRs?

packages/app/src/features/dialogs/common/components/TokenValueChange.tsx
yivlad: I think it is not as important for supply or borrow dialogs

PR #152, opened 2024-07-04
Add more assertions to native withdraw hooks

oskarvu:
Closes FRO-780


yivlad (approved):

packages/app/src/domain/psm-actions/useWithdrawAndSwap.ts
```diff
@@ -29,15 +33,19 @@ export function useWithdrawAndSwap({
...
+  mode,
+  reserveAddresses,
   onTransactionSettled,
   enabled = true,
 }: UseWithdrawAndSwapArgs): ReturnType<typeof useWrite> {
   const client = useQueryClient()
   const chainId = useChainId()
+  const { address: owner } = useAccount()
+
+  assertNativeWithdraw({ mode, receiver: _receiver, owner, reserveAddresses })
```
krzkaczor, line 45: It would be much better to run these assertions right before writing to the blockchain. So we need a callback in `useWrite` like `onTransactionSend`. Then we can tweak assertions so they are even more stict and require that receiver is not undefined (because at that point we will be sure about it).

packages/app/src/domain/hooks/useWrite.ts
```diff
@@ -25,6 +25,7 @@ export interface UseWriteResult {
 
 export interface UseWriteCallbacks {
   onTransactionSettled?: () => void
+  withdrawReceiverSanityCheck?: () => void
```
krzkaczor, line 28: Why to give it a very specific name? `useWrite` doesn't know anything about the domain. I would simply name it `onBeforeWrite` and add a comment that this is a place where extra sanity checks can happen.

krzkaczor, line 28:
```suggestion
  onBeforeWrite?: () => void // @note: good place to run extra sanity checks
```

yivlad, line 28: I think this could be a generic callback like `onTransactionSend`

krzkaczor (approved):

PR #153, opened 2024-07-04
Tenderly vnets support

krzkaczor:


packages/app/src/features/actions/ActionsContainer.PageObject.ts
```diff
@@ -27,11 +27,15 @@ export class ActionsPageObject extends BasePageObject {
...
   // #region actions
-  async acceptAllActionsAction(expectedNumberOfActions: number): Promise<void> {
+  async acceptAllActionsAction(expectedNumberOfActions: number, forkContext?: ForkContext): Promise<void> {
     for (let index = 0; index < expectedNumberOfActions; index++) {
       const row = this.region.getByTestId(testIds.actions.row(index))
 
       await row.getByRole('button', { name: actionButtonRegex }).click()
+      if (forkContext?.isVnet) {
+        await expect(row.getByRole('button', { name: actionButtonRegex })).not.toBeVisible()
+        await forkContext.progressSimulation(this.page, 5)
```
yivlad, line 37: I think this should be before the click, otherwise we are introducing race condition with what request is sent first, tx or `evm_setNextBlockTimestamp`

packages/app/src/config/consts.ts
```diff
@@ -14,3 +14,6 @@ export const ZUSTAND_APP_STORE_LOCAL_STORAGE_VERSION = 2
 
 export const apiUrl = import.meta.env.VITE_API_URL ?? '/api'
 export const blockAnaliticaApiUrl = import.meta.env.VITE_BLOCK_ANALITICA_API_URL ?? '/ba-api'
+
+// @note: all sandboxes created by the app begin (when expressed as strings) with this chain id. Ex: 30301719211032
+export const SANDBOX_NETWORKS_CHAIN_ID_PREFIX = 3030
```
yivlad, line 19: Why is this extracted form zustand store?

krzkaczor: I think this is not something that can change in the future so i thought we should treat it as any other constant.

packages/app/src/features/actions/ActionsContainer.PageObject.ts
krzkaczor: Actually we are not introducing a race condition because TX is always sent first and mined first (click + waiting for button to dissapear).

`evm_setNextBlockTimestamp` works in a way that it changes timestamp of the next block. We set this when network is created first and then after any tx we need to set it again. That's what's happening in this code.

packages/app/src/domain/tenderly/TenderlyVnetClient.ts
```diff
@@ -0,0 +1,57 @@
+import { solidFetch } from '@/utils/solidFetch'
+
+export interface CreateTenderlyForkArgs {
```
oskarvu, line 3: `CreateTenderlyVnetArgs`

packages/app/src/pages/Root.tsx
```diff
@@ -17,6 +18,7 @@ export function RootRoute() {
           <DialogDispatcherContainer />
           <ComplianceContainer />
         </ErrorBoundary>
+        {import.meta.env.VITE_DEV_DEBUG === 1 && <Debug />}
```
oskarvu, line 21: Would be nice to add it to the env example and maybe other env files.

packages/app/src/test/e2e/setup.ts
```diff
@@ -119,22 +119,45 @@ export async function injectFunds(
...
-        forkContext.forkUrl,
-        (TOKENS_ON_FORK as any)[forkContext.chainId][tokenName].address,
-        address,
-        BaseUnitNumber(
-          parseUnits(balance.toString(), (TOKENS_ON_FORK as any)[forkContext.chainId][tokenName].decimals),
-        ),
-      )
+  if (forkContext.isVnet) {
+    const promises = Object.entries(assetBalances).map(async ([tokenName, balance]) => {
+      if (tokenName === 'ETH' || tokenName === 'XDAI') {
```
oskarvu, line 124: Maybe worth to extract and reuse it? It's same for vnets and forks.

krzkaczor: i hope this code will be removed soonish so i woudn't care

oskarvu (approved):

packages/app/src/features/actions/ActionsContainer.PageObject.ts
yivlad: Ah, makes sense. I guess then we should leave a comment that here we're setting things up for the next transaction. Also it's kinda confusing in the `progressSimulation` function itself:
```ts
await injectUpdatedDate(page, this.simulationDate)
await tenderlyRpcActions.evmSetNextBlockTimestamp(forkContext.forkUrl, Number(newTimestamp))
```
If I understand correctly here we have the timestamp of the future block in the browser but on the vnet the previous timestamp is the current one? I think this is generally possible situation but misleading anyway.

PR #154, opened 2024-07-05
feat: 🎸 sum market details wallet asset with its native equiv

EmilMalanczak:


packages/app/src/pages/MarketDetails.test-e2e.ts
```diff
@@ -195,6 +195,61 @@ test.describe('Market details', () => {
...
+      })
+
+      const borrowPage = new BorrowPageObject(page)
+      await borrowPage.depositWithoutBorrowActions({ ...initialDeposits })
+
+      await page.goto(buildUrl('marketDetails', { asset: WETH, chainId: fork.chainId.toString() }))
+
+      const marketDetailsPage = new MarketDetailsPageObject(page)
+
+      await marketDetailsPage.expectWalletBalance('15.00 WETH')
```
yivlad, line 251: We can skip depositing here as we do not need the borrow button to be active. I'd set the `initialPage` in setup to market-details and just do the assertion

packages/app/src/pages/MarketDetails.test-e2e.ts
```diff
@@ -195,6 +195,61 @@ test.describe('Market details', () => {
       await borrowDialog.expectDialogHeader('Borrow WETH')
       await borrowDialog.closeDialog()
     })
+
+    test('opens dialogs for WETH when having only ETH', async ({ page }) => {
```
krzkaczor, line 199: even though this is the current behaviour and fixing it is out of scope I would leave a todo note that probably in this scenario, dialog should realize that user has only ETH and should open on ETH tab. Because right now we force users to switch manually. (basically an asset with the biggest balance should be chosen automatically).

yivlad (approved):

oskarvu (approved):

PR #164, opened 2024-07-22
Add future market calculations to withdraw MAX logic

yivlad:
- Adds approve  delegation value when withdrawing native asset calculation based on future market state.
- Adds tests for setting approve delegation value when withdrawing native asset.
- Adds `expectTransactionToBeSent` function. Can't be used now due to inability to set deterministic timestamps for transactions in E2E tests.


packages/app/src/features/actions/flavours/withdraw/types.ts
```diff
@@ -7,6 +7,7 @@ export interface WithdrawObjective {
   reserve: Reserve
   value: NormalizedUnitNumber
   all: boolean
+  gatewayApprovalValue: NormalizedUnitNumber
```
krzkaczor, line 10: imo this shoudl be optional and undefined should be passed if not dealing with native asset (anyway, this abstraction leaks ;/ ).

packages/app/src/features/dialogs/withdraw/WithdrawDialog.test-e2e.ts
```diff
@@ -601,5 +601,44 @@ test.describe('Withdraw dialog', () => {
...
+      await withdrawDialog.closeDialog()
+
+      await dashboardPage.clickDepositButtonAction('WETH')
+      await depositDialog.selectAssetAction('ETH')
+      await depositDialog.fillAmountAction(4)
+      await depositDialog.actionsContainer.acceptAllActionsAction(1)
+      await depositDialog.viewInDashboardAction()
+      await dashboardPage.expectDepositedAssets(33_530)
+
+      // following checks leverage the fact that approval is cached, therefore we input different values to estimate the approval value
```
krzkaczor, line 633: uhh we should simply have a way to assert whaterver tx was signed.... any idea how to implement this? Maybe there is a hook in viem wallet configuration? Then we would simply write to some kind of global data structure and read it from withing playwright runner?

yivlad: This can be done if we catch calls to RPC endpoint and look for `eth_getTransactionByHash` method. Do we want to do this though? Generally this mechanism can be useful,  but here it seems to be an overkill.

packages/app/src/domain/hooks/useWrite.ts
```diff
@@ -78,6 +79,13 @@ export function useWrite<TAbi extends Abi, TFunctionName extends ContractFunctio
   useEffect(() => {
     if (txReceipt) {
       callbacks.onTransactionSettled?.()
+
+      if (import.meta.env.VITE_PLAYWRIGHT === '1' && parameters?.request) {
```
krzkaczor, line 83: please leave a comment explaining why we need this. Can we extract everything between {...} to separate function? God this code is ugly 😆

oskarvu: Make sure to be consistent with other uses of `VITE_PLAYWRIGHT`, so it's always compared like this.

oskarvu (approved):

packages/app/src/domain/hooks/useWrite.ts
```diff
@@ -78,6 +79,10 @@ export function useWrite<TAbi extends Abi, TFunctionName extends ContractFunctio
   useEffect(() => {
     if (txReceipt) {
       callbacks.onTransactionSettled?.()
+
```
krzkaczor, line 82:
```suggestion
// for e2e tests needs we record sent txs
```

krzkaczor (approved):

PR #169, opened 2024-07-24
Increase sentry max value length

yivlad:
All errors in sentry were trimmed to default `maxValueLength` parameter in Sentry - 250. This makes it impossible to fully read contract function execution errors, as they are oftentimes longer than 250 characters. This PR increases the `maxValueLength` to 5000.


krzkaczor (approved):
💯

oskarvu (approved):

PR #171, opened 2024-07-25
Introduce useTokens hook

yivlad:
Closes FRO-804


packages/app/src/config/chain/index.ts
```diff
@@ -74,6 +76,11 @@ const chainConfig: ChainConfig = {
       [TokenSymbol('USDC'), TokenSymbol('sDAI')],
     ],
     savingsNativeRouteTokens: [TokenSymbol('DAI'), TokenSymbol('USDC')],
+    extraTokens: [
+      { address: CheckedAddress('0x6b175474e89094c44da98b954eedeac495271d0f'), oracleType: 'fixed-usd' }, // DAI
+      { address: CheckedAddress('0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'), oracleType: 'fixed-usd' }, // USDC
+      { address: CheckedAddress('0x83f20f44975d03b1b09e64809b757c47f942beea'), oracleType: 'erc4626' }, // sDAI
```
krzkaczor, line 82: Hooks use simply 'vault' instead of 'erc4626'. I would suggest using the same name in both places. (i like vault tbh)...

krzkaczor (approved):
I see that you decided against relying on pot data to calculate price. I think it's fine but let's keep this in mind that maybe we will have to rethink it once we have new pot data...

PR #173, opened 2024-07-26
Basic analytics

krzkaczor:
Basic analytics using plausible. 

Closes FRO-810

### DIFF
--- a/packages/app/src/domain/hooks/useWalletType.ts
+++ b/packages/app/src/domain/hooks/useWalletType.ts
@@ -1,10 +1,12 @@
 import { useAccount } from 'wagmi'
+import { useSandboxState } from '../sandbox/useSandboxState'
 import { useIsSmartContract } from './useIsSmartContract'
 
-export type WalletType = 'gnosis-safe' | 'universal'
+export type WalletType = 'gnosis-safe' | 'sandbox' | 'unknown' | string // string means connector type ie. 'walletconnect' or 'metamask'
 
 export function useWalletType(): WalletType | undefined {
   const { address, connector } = useAccount()
+  const { isInSandbox } = useSandboxState()
   const canBeGnosisSafe = connector?.name === 'WalletConnect' || connector?.name === 'Safe' // avoids querying bytecode if not needed
   const { isSmartContract } = useIsSmartContract(canBeGnosisSafe ? address : undefined)
 
@@ -13,7 +15,10 @@ export function useWalletType(): WalletType | undefined {
   }
 
   if (!canBeGnosisSafe) {
-    return 'universal'
+    if (isInSandbox) {
+      return 'sandbox'
+    }
+    return connector?.name || 'unknown'
   }
 
   if (isSmartContract) {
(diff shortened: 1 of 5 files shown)


packages/app/src/domain/hooks/useWalletType.ts
```diff
@@ -1,10 +1,12 @@
...
 import { useIsSmartContract } from './useIsSmartContract'
 
-export type WalletType = 'gnosis-safe' | 'universal'
+export type WalletType = 'gnosis-safe' | 'sandbox' | 'unknown' | string // string means connector type ie. 'walletconnect' or 'metamask'
 
 export function useWalletType(): WalletType | undefined {
   const { address, connector } = useAccount()
+  const { isInSandbox } = useSandboxState()
   const canBeGnosisSafe = connector?.name === 'WalletConnect' || connector?.name === 'Safe' // avoids querying bytecode if not needed
   const { isSmartContract } = useIsSmartContract(canBeGnosisSafe ? address : undefined)
```
krzkaczor, line 11: btw. iam curious if you think this hook needs memo or something? It seems to be working totally fine because getting bytecode is cached by react-query (wagmi)

yivlad: I think it's fine as it is

oskarvu: I don't think we need to memoise it.

packages/app/src/domain/hooks/useWalletType.ts
```diff
@@ -1,10 +1,12 @@
 import { useAccount } from 'wagmi'
+import { useSandboxState } from '../sandbox/useSandboxState'
 import { useIsSmartContract } from './useIsSmartContract'
 
-export type WalletType = 'gnosis-safe' | 'universal'
+export type WalletType = 'gnosis-safe' | 'sandbox' | 'unknown' | string // string means connector type ie. 'walletconnect' or 'metamask'
```
yivlad, line 5: Type union with `string` doesn't make sense as it produces only string. I guess we have to return string fro the hook

oskarvu (approved):

packages/app/src/domain/hooks/useWrite.ts
```diff
@@ -123,6 +126,14 @@ export function useWrite<TAbi extends Abi, TFunctionName extends ContractFunctio
           sanityCheckTx(parameters.request, chainId)
           callbacks.onBeforeWrite?.()
 
+          recordEvent('tx-sent', {
```
yivlad, line 129: Can we differentiate between transactions sent in sandbox mode and other ones?

krzkaczor: no need, we have `walletType` prop for that

packages/app/src/domain/hooks/useWalletType.ts
krzkaczor: I know but this is for documentation purpose...

yivlad (approved):

PR #177, opened 2024-07-30
Create basic new savings view

oskarvu:
Closes FRO-806


packages/app/src/features/savings-with-nst/views/SavingsView.tsx
```diff
@@ -1,88 +1,198 @@
...
   assetsInWallet,
   totalEligibleCashUSD,
   maxBalanceToken,
   openDialog,
-  nstAPY,
+  ...savingTokensDetails
 }: SavingsViewProps) {
-  const displaySavingsDai = depositedUSD.gt(0)
-  const displaySavingsOpportunity = opportunityProjections.thirtyDays.gt(0) && opportunityProjections.oneYear.gt(0)
+  if ('sDai' in savingTokensDetails && 'sNST' in savingTokensDetails) {
```
yivlad, line 34: I think this routing should be in the container, not in the view

packages/app/src/features/savings-with-nst/logic/useSavings.ts
```diff
@@ -6,112 +7,141 @@ import { useSavingsNstInfo } from '@/domain/savings-info/useSavingsNstInfo'
...
         maxBalanceToken: TokenWithBalance
-        chainId: SupportedChainId
-      }
-    | { state: 'unsupported' }
-  savingsNSTDetails:
-    | {
-        state: 'supported'
-        APY: Percentage
-      }
+        originChainId: SupportedChainId
```
yivlad, line 39: Why origin and not just chainId?

packages/app/src/features/savings-with-nst/logic/useSavings.ts
```diff
@@ -6,112 +7,141 @@ import { useSavingsNstInfo } from '@/domain/savings-info/useSavingsNstInfo'
...
+  depositedUSDPrecision: number
+}
+
 export interface UseSavingsResults {
   guestMode: boolean
   openDialog: OpenDialogFunction
   openSandboxModal: () => void
-  savingsDaiDetails:
-    | {
+  savingsDetails:
```
yivlad, line 33: My idea was to make a clear distinction between savings DAI and savings NST. These two entities aren't actually connected in any way, so i think the separation makes sense. I think the only benefit in the current implementation is that we get to share common fields like  `assetsInWallet`? Looks like not a big deal to me. Can we explore the way with creating separate fields `savingsDaiDetails` and `savingsNSTDetails`? It will make the structure of `savingsDetails` much flatter and imo easier to work with. Open for discussion here.

packages/app/src/features/savings-with-nst/logic/useSavings.ts
oskarvu: This origin chain id is used in view component, to display different text based on chain. I suppose I could return just chain id and then get config (to get chain id) directly in component.

It was like this before, I changed name to `originChainId` to make it more clear, what is returned, and so I can use `chainId` inside hook.

packages/app/src/features/savings-with-nst/logic/useSavings.ts
oskarvu: I'm not convinced that we should separate it. For me it makes clear, what parts of savings details are shared, and what parts are distinct. I think it also nicely describes possible states (that you can have either dai, nst, or both). Having separate `savingsDaiDetails` and `savingsNSTDetails` will require to make them both "optional".

packages/app/src/domain/savings-info/useSavingsDaiInfo.ts
```diff
@@ -15,14 +13,12 @@ export function useSavingsDaiInfo(): UseSavingsDaiInfoResult {
...
+
   const result = useSuspenseQuery(
-    queryOptions({
-      wagmiConfig,
-      chainId,
-      timestamp,
-    }),
+    queryOptions
+      ? queryOptions({ wagmiConfig, chainId, timestamp })
+      : { queryKey: ['savings-info-unsupported'], queryFn: () => null },
```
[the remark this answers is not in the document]
oskarvu: When I tested it, it worked, but still a good idea to make it more descriptive/specific.

packages/app/src/features/savings-with-nst/views/SavingsView.tsx
oskarvu: Yeah, I think you're right. Then I need to make separate views and separate stories for each of that views.

packages/app/src/features/navbar/logic/useNavbarSavingsInfo.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (savingsNstInfoQuery) return savingsNstInfoQuery
+    if (savingsDaiInfoQuery) return savingsDaiInfoQuery
+    return undefined
+  }
+  const savingsInfoQueryOptions = getSavingsInfoQuery()
+
+  const queryResult = useQuery(
+    savingsInfoQueryOptions
+      ? savingsInfoQueryOptions({ wagmiConfig, chainId, timestamp })
+      : { queryKey: ['navbar-unsupported-savings-info'], queryFn: skipToken },
```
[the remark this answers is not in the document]
oskarvu: I guess not, but it's more for a documentation purpose.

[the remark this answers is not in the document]
krzkaczor: I think it's cleaner to skip it entirely. Usage of `skipToken` makes it clear IMO.

krzkaczor (approved):

yivlad (approved):

PR #178, opened 2024-07-30
Invalidate allowance query for PSM actions hook

yivlad:


krzkaczor (approved):
Good abstraction 👍

oskarvu (approved):
Nice 🥇

PR #180, opened 2024-07-31
Add NST to sNST deposit support

yivlad:


packages/app/src/features/actions/flavours/native-sdai-deposit/maker-stables/useCreateMakerStableToSavingsHandler.ts
```diff
@@ -1,22 +1,22 @@
...
-export interface UseCreateDaiToSDaiDepositHandlerOptions {
+export interface UseCreateMakerStableToSavingsHandlerOptions {
   enabled: boolean
   onFinish?: () => void
 }
 
-export function useCreateDaiToSDaiDepositHandler(
-  action: DaiToSDaiDepositAction,
-  options: UseCreateDaiToSDaiDepositHandlerOptions,
+export function useCreateMakerStableToSavingsHandler(
```
krzkaczor, line 11: IMO this is all fine. We have one action that is configurable and takes two tokens. Here we will have a bunch of if statements and we will write tests for it that will call `onAction` and expect that the correct hook was called. I think it's gonna work fine.

krzkaczor (approved):

oskarvu (approved):
I think we need to introduce updated approach of objective -> action flow.

PR #181, opened 2024-08-01
Add migrate from sDai to NST (withdraw & redeem) hooks

oskarvu:
Closes FRO-834
Closes FRO-835

### DIFF
--- a/packages/app/src/domain/migration-actions/useMigrateSDAISharesToNST.ts
+++ b/packages/app/src/domain/migration-actions/useMigrateSDAISharesToNST.ts
@@ -1,17 +1,19 @@
+import { migrationActionsAbi } from '@/config/abis/migrationActionsAbi'
+import { MIGRATE_ACTIONS_ADDRESS } from '@/config/consts'
 import { Mode } from '@/features/dialogs/savings/withdraw/types'
 import { toBigInt } from '@/utils/bigNumber'
 import { useQueryClient } from '@tanstack/react-query'
-import { erc4626Abi } from 'viem'
 import { useAccount, useChainId } from 'wagmi'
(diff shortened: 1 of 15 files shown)


packages/app/src/domain/migration-actions/useMigrateSDAISharesToNST.ts
```diff
@@ -0,0 +1,72 @@
...
+  reserveAddresses,
+  onTransactionSettled,
+  enabled = true,
+}: UseMigrateSDAISharesToNSTArgs): ReturnType<typeof useWrite> {
+  const client = useQueryClient()
+  const chainId = useChainId()
+  const { address: owner } = useAccount()
+  const receiver = _receiver || owner
+
+  const config = ensureConfigTypes({
```
krzkaczor, line 39: is using `ensureConfigTypes` needed here? i think it was only needed when doing something conditionally.

yivlad: I think what @krzkaczor meant was that if it is not conditional, it can be inlined. The problem with current approach that if something goes wrong we'll get type error not in the place where the object is declared, but where the object is used. It can be problematic especially if we return the config from function. Therefore I there should be 2 approaches:
1. Using `ensureConfig` when we declare standalone object
2. Inlining the config into the `useWrite` hook.

But as for me `ensureConfig` is fine even if it's not generated conditionally. The most important thing is to have typesafety right where we declare the config object.

oskarvu: Inlined config in `useWrite` hook.

krzkaczor (approved):

yivlad (approved):

PR #183, opened 2024-08-01
Add useMigrateSDAISharesToSNST hook

yivlad:


packages/app/src/domain/migration-actions/useMigrateSDAISharesToSNST.ts
```diff
@@ -0,0 +1,68 @@
...
+import { allowanceQueryKey } from '../market-operations/allowance/query'
+import { assertNativeWithdraw } from '../savings/assertNativeWithdraw'
+import { CheckedAddress } from '../types/CheckedAddress'
+import { BaseUnitNumber } from '../types/NumericValues'
+import { getBalancesQueryKeyPrefix } from '../wallet/getBalancesQueryKeyPrefix'
+
+export interface useMigrateSDAISharesToSNSTArgs {
+  sDai: CheckedAddress
+  sDaiAmount: BaseUnitNumber
+  receiver?: CheckedAddress
```
oskarvu, line 17: Will we have a use case where we send `sNST` to someone? I think this hook would be only used for upgrading (migrating) from `sDai` to `sNST`, right?

krzkaczor: right 👍

oskarvu (approved):

PR #188, opened 2024-08-06
Prevent risk acknowledgment from flashing at input changes

oskarvu:
Closes FRO-839


krzkaczor (approved):
This looks good, do other dialogs need similiar change? (maybe just create a ticket for those) 

Also, this prompted me to create another ticket about modal size flickering: https://linear.app/phoenixlabs/issue/FRO-857/to-reduce-flickering-default-tx-details-should-be-always-shown-on-the

oskarvu:
@krzkaczor Other dialogs that display liquidation risk acknowledgment don't have debouncing, so the problem is not present.

krzkaczor (approved):

yivlad (approved):

PR #191, opened 2024-08-06
Refactor actions to use action configs

yivlad:
Closes FRO-855


packages/app/src/utils/useOnDepsChange.ts
```diff
@@ -0,0 +1,12 @@
+import { DependencyList, useRef } from 'react'
+
+// unlike useEffect, this hook is guaranteed to run the callback even if the deps change fast
```
krzkaczor, line 3:
```suggestion
// unlike useEffect, this hook is guaranteed to run the callback even if the deps change fast (think a render that triggers another render by changing the state for example)
```

packages/app/src/domain/hooks/useWrite.ts
```diff
@@ -169,3 +177,29 @@ function storeRequest(request: any): void {
   txList.push({ ...request, calldata })
   window[__TX_LIST_KEY] = txList as any
 }
+
+function useWriteContract(
```
krzkaczor, line 181: we need a comment explaining difference between this and what is exported in wagmi

krzkaczor (commented):
There is a regression in this PR:
1. go to dashboard (in sandbox mode)
2. Deposit `sDAI` -> 1000 sdai
3. Click permit (doesn't work with approvals as well)
4. Change to 1001 sDAI
5. process doesn't revert and tx simulation fails.

We need a E2E test for this and fix. Let's do it as another PR targeting this branch.

packages/app/src/features/actions/logic/useCreateActions.ts
```diff
@@ -117,13 +130,26 @@ export function useCreateActions(objectives: Objective[]): Action[] {
           value: objective.value,
           useAToken: objective.useAToken,
         }
+
         if (objective.reserve.token.symbol === nativeAssetInfo.nativeAssetSymbol || objective.useAToken) {
           return [repayAction]
         }
+
+        if (actionsSettings.preferPermits && chainConfig.permitSupport[objective.reserve.token.address]) {
```
krzkaczor, line 138: nit: WDYT about supporting permits this way: or even maybe create a function that takes care of creating pemit or approve based on the flag... food for thought. Might be refactored later

```typescript
        const shouldUsePermit =
          actionsSettings.preferPermits && chainConfig.permitSupport[objective.reserve.token.address]

        const approveAction: PermitAction | ApproveAction = shouldUsePermit
          ? {
              type: 'permit',
              token: objective.reserve.token,
              spender: lendingPool,
              value: objective.value,
            }
          : {
              type: 'approve',
              token: objective.reserve.token,
              spender: lendingPool,
              requiredValue: objective.requiredApproval,
              value: objective.value,
            }
        return [approveAction, repayAction]
```


IMO this reads better...

krzkaczor (approved):

PR #235, opened 2024-08-29
Add market details links

EmilMalanczak:


EmilMalanczak:
I was confused about the `aToken` symbol because of the difference between our application and etherscan results on a specific address.

As I discussed with @yivlad, it would be good to rename `aToken` (aave token) to `spToken` (spark token).

what are your thoughts @krzkaczor ?

yivlad (approved):

krzkaczor (commented):
This is almost good. There is this edge case when user switches the network and we display market details for mainnet (chaid is part of the url) but links are pointing to new network (ie gnosis). 

I think a simple solution would be for `useBlockExplorerAddressLink` to receipt an optional `chainId` argument that overrides currently selected network. 

https://github.com/user-attachments/assets/9d7886a8-2a49-45af-a2c3-c34230a07a0f

krzkaczor (approved):

PR #239, opened 2024-08-30
Display cap automator info on market details

EmilMalanczak:


packages/app/src/features/market-details/views/components/FullView.tsx
```diff
@@ -34,11 +35,13 @@ export function FullView({
...
         <div className="flex flex-col gap-6">
-          {marketOverview.supply && <SupplyStatusPanel token={token} {...marketOverview.supply} />}
+          {marketOverview.supply && (
+            <SupplyStatusPanel token={token} {...marketOverview.supply} capInfo={capAutomatorInfo.supplyCap} />
+          )}
           {marketOverview.lend && <LendStatusPanel {...marketOverview.lend} />}
           <CollateralStatusPanel {...marketOverview.collateral} />
           {marketOverview.eMode && <EModeStatusPanel {...marketOverview.eMode} />}
-          <BorrowStatusPanel token={token} {...marketOverview.borrow} />
+          <BorrowStatusPanel token={token} {...marketOverview.borrow} capInfo={capAutomatorInfo.borrowCap} />
```
yivlad, line 44: I think `capInfo` logically fits into the `marketOverview`

packages/app/src/features/market-details/components/status-panel/BorrowStatusPanel.tsx
```diff
@@ -78,6 +82,32 @@ export function BorrowStatusPanel({
...
+            <InfoTile>
+              <InfoTile.Label>Instantly available borrow cap:</InfoTile.Label>
+              <InfoTile.Value>
+                {token.format(borrowCap!, { style: 'compact' })} {token.symbol}
+                <CooldownTimer
+                  renewalPeriod={capInfo.increaseCooldown}
+                  latestUpdateTimestamp={capInfo.lastIncreaseTime}
+                />
+              </InfoTile.Value>
+              <InfoTile.ComplementaryLine>{token.formatUSD(borrowCap!, { compact: true })}</InfoTile.ComplementaryLine>
```
yivlad, line 107: I guess we can also put `borrowCap` check near above to avoid using `!`

krzkaczor (commented):
image

this is not aligned

packages/app/src/features/market-details/components/status-panel/BorrowStatusPanel.tsx
```diff
@@ -86,3 +86,43 @@ export function BorrowStatusPanel({
     </Panel.Wrapper>
   )
 }
+
+interface capAutomatorInfoTileProps {
```
yivlad, line 90:
```suggestion
interface CapAutomatorInfoTileProps {
```

packages/app/src/features/market-details/components/status-panel/BorrowStatusPanel.tsx
```diff
@@ -86,3 +86,43 @@ export function BorrowStatusPanel({
...
+  token: Token
+  capAutomatorInfo: CapConfig | null
```
krzkaczor, line 92: Generally speaking we avoid using nulls. `undefined` values are enough in TypeScript as long as you don't need to differentiate between something loading (undefined) vs something empty (null). This is not a case here. Basically we only use `null` when dealing with before-mentioned case or returning stuff from react-query.

So:
```
capAutomatorInfo?: CapConfig
```

Furthermore let's get rid of nulls in `CapAutomatorInfo` interface so it becomes:
```
export interface CapAutomatorInfo {
  supplyCap?: CapConfig
  borrowCap?: CapConfig
}
```
(it's suspended so we don't care about loading state in types).

packages/app/src/features/market-details/components/status-panel/BorrowStatusPanel.tsx
```diff
@@ -86,3 +86,43 @@ export function BorrowStatusPanel({
...
+  borrowCap?: NormalizedUnitNumber
+}
+
+function CapAutomatoInfoTile({ token, capAutomatorInfo, borrowCap }: capAutomatorInfoTileProps) {
```
yivlad, line 96: Should we just return `undefined` from the component when `borrowCap` is `undefined`? To avoid empty divs, I guess empty divs can break the grid

packages/app/src/features/market-details/logic/makeDaiMarketOverview.ts
```diff
@@ -12,9 +12,23 @@ export interface MakeDaiMarketOverviewParams {
...
+    },
+  })
   const sDAI = marketInfo.findOneReserveByToken(marketInfo.sDAI)
-  const sDaiOverview = makeMarketOverview({ reserve: sDAI, marketInfo })
+  const sDaiOverview = makeMarketOverview({
+    reserve: sDAI,
+    marketInfo,
+    capAutomatorInfo: {
+      borrowCap: null,
+      supplyCap: null,
```
yivlad, line 29: Can you verify that sDAI supply indeed does not have cap automator? Technically DAI and sDAI are different markets

yivlad:
```sh
cast call -r https://eth-mainnet.g.alchemy.com/v2/WVOCPHOxAVE1R9PySEqcO7WX2b9_V-9L 0x2276f52afba7Cf2525fd0a050DF464AC8532d0ef "function supplyCapConfigs(address) public view returns (uint48, uint48, uint48, uint48, uint48)" 0x83f20f44975d03b1b09e64809b757c47f942beea
```

packages/app/src/features/market-details/components/status-panel/BorrowStatusPanel.tsx
oskarvu: I would assume that data is available in this component, and render it conditionally at `BorrowStatusPanel`. Similarly in `SupplyStatusPanel`.

yivlad (approved):

packages/app/src/locales/en.po
```diff
@@ -6,6 +6,12 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: @lingui/cli\n"
 "Language: en\n"
+"Project-Id-Version: \n"
```
oskarvu, line 9: Where this change come from?

EmilMalanczak: it came with the previous PR - basically we decided to ignore `*.po` files but after putting them into `.gitinore` it caused CI to fail on the test step as there is no build step which creates these files. So we ignored `*.po` excluding `en.po` which we treat like a standard file - it was changed while I checked if the code is building properly

packages/app/src/features/market-details/types.ts
```diff
@@ -8,13 +8,25 @@ import { NormalizedUnitNumber, Percentage } from '@/domain/types/NumericValues'
 import { Token } from '@/domain/types/Token'
 import { TokenSymbol } from '@/domain/types/TokenSymbol'
 
+import { CapConfig } from '@/domain/cap-automator/types'
```
krzkaczor, line 11: One final comment, I just realized that `CapConfig` should be renamed to `CapAutomatorConfig`

krzkaczor (approved):

oskarvu (approved):

PR #243, opened 2024-09-03
Show apy benefit only when applicable

oskarvu:
Closes FRO-930


packages/app/src/features/dialogs/migrate/common/logic/useMigrateDialog.ts
```diff
@@ -48,7 +47,7 @@ export function useMigrateDialog({ type, fromToken, toToken }: UseMigrateDialogP
   return {
     objectives,
     migrationAmount: fromTokenBalance,
-    apyDifference,
+    apyImprovement: apyDifference.gt(0) ? apyDifference : undefined,
```
krzkaczor, line 50: `apyDifference` is an absolute value so it can't be bigger than 0 👀

yivlad: I think the point of this PR should be removing this constraint. APY can generally be negative as well, so I think we should add some kind of flag to `Percentage` type that enables it to be negative. This will make negative APY difference a valid scenario as well.

oskarvu: I forgot to remove absolute value from `apyDifference` in this hook like I did in `makeMigrationInfo`. Let me fix that.

The plan is to return `apyImprovement` only when `apyDifference` is greater than zero and then render benefit accordingly, as ticket states.

yivlad (approved):

EmilMalanczak (approved):

PR #259, opened 2024-09-10
Savings improvements from feedback

EmilMalanczak:


packages/app/src/features/savings/logic/useSavings.ts
```diff
@@ -122,10 +122,12 @@ export function useSavings(): UseSavingsResults {
...
-  const assetsInWallet = inputTokens.map((tokenWithBalance) => ({
-    ...tokenWithBalance,
-    blockExplorerLink: getBlockExplorerLink(tokenWithBalance.token.address),
-  }))
+  const assetsInWallet = inputTokens
+    .map((tokenWithBalance) => ({
+      ...tokenWithBalance,
+      blockExplorerLink: getBlockExplorerLink(tokenWithBalance.token.address),
+    }))
+    .sort((a, b) => b.balance.comparedTo(a.balance))
```
krzkaczor, line 130: if balances are the same, can we prioriteze USDS to be on the top of the list?

packages/app/src/features/savings/components/cash-in-wallet/components/MoreDropdown.tsx
```diff
@@ -38,7 +40,7 @@ export function MoreDropdown({ token, blockExplorerLink, migrationInfo, disabled
         <LinkDecorator to={blockExplorerLink} external>
           <DropdownItem>
             <DocumentSketchIcon className="h-4 w-4" />
-            Learn more
+            View contract
             <ArrowUpRightIcon className="ml-auto h-4 w-4" />
```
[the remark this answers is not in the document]
EmilMalanczak: Replaced the icon and removed the old one as it wasn't used anymore in the app

packages/app/src/features/savings/logic/useSavings.ts
```diff
@@ -122,10 +122,23 @@ export function useSavings(): UseSavingsResults {
...
+      blockExplorerLink: getBlockExplorerLink(tokenWithBalance.token.address),
+    }))
+    .sort((a, b) => {
+      const balanceComparison = b.balance.comparedTo(a.balance)
+
+      // Sort by balance
+      if (balanceComparison !== 0) return balanceComparison
+
+      // Prioritize token with symbol 'usds'
+      if (a.token.symbol === 'usds') return -1
```
krzkaczor, line 137: i think this should use USDS symbol from network config. Also it doesn't work in case of empty balances:

image

packages/app/src/ui/atoms/button/Button.tsx
```diff
@@ -10,7 +10,7 @@ const buttonVariants = cva(
   {
     variants: {
       variant: {
-        primary: 'bg-primary-bg text-primary-foreground hover:bg-primary-hover',
+        primary: 'bg-primary-bg text-primary-foreground hover:bg-primary-hover hover:text-primary-foreground',
```
EmilMalanczak, line 13: Fixes FRO-932 

Issue appear on link buttons as underlying Link component was attaching hover text color class which was overriding Button color

krzkaczor: This is OK fix but it's not visible in chromatic meaning that button is missing some stories. Can you add them?

krzkaczor (approved):

oskarvu (approved):

PR #261, opened 2024-09-11
Update ToS 

EmilMalanczak:


packages/app/src/features/compliance/ComplianceContainer.tsx
```diff
@@ -12,7 +12,13 @@ export function ComplianceContainer() {
 
   return (
     <Dialog open={visibleModal.type !== 'none'}>
-      <DialogContent showCloseButton={false}>
+      <DialogContent
+        showCloseButton={false}
+        // https://github.com/radix-ui/primitives/issues/2248
```
krzkaczor, line 17: Linking to issue is good but also we should try to describe the issue in a sentence like:
```suggestion
        // Prevents tooltip open automatically https://github.com/radix-ui/primitives/issues/2248
```

packages/app/src/features/compliance/components/TermsOfService.tsx
```diff
@@ -66,6 +68,20 @@ export interface TermsOfServiceProps {
 }
 
 export function TermsOfService({ onAgree }: TermsOfServiceProps) {
+  const [hasScrolled, setHasScrolled] = useState(false)
```
krzkaczor, line 71: Let's rename it to something like: `hasUserReadTerms`

krzkaczor (approved):

PR #262, opened 2024-09-11
Fro 948 upgrade downgrade inputs

oskarvu:


packages/app/src/features/dialogs/migrate/common/logic/form.ts
```diff
@@ -0,0 +1,31 @@
...
+import { AssetInputSchema } from '@/features/dialogs/common/logic/form'
+import { FormFieldsForDialog } from '@/features/dialogs/common/types'
+import { UseFormReturn } from 'react-hook-form'
+
+export function getFormFieldsForMigrateDialog(
+  form: UseFormReturn<AssetInputSchema>,
+  tokensInfo: TokensInfo,
+): FormFieldsForDialog {
+  // eslint-disable-next-line func-style
+  const changeAsset = (newSymbol: TokenSymbol): void => {
```
krzkaczor, line 12: this doesn't seem to be needed since we don't allow to change asset on the form.

packages/app/src/features/dialogs/migrate/upgrade/UpgradeDialog.tsx
```diff
@@ -11,7 +11,12 @@ interface UpgradeDialogProps extends CommonDialogProps {
 export function UpgradeDialog({ fromToken, toToken, open, setOpen }: UpgradeDialogProps) {
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      <DialogContent className="p-0">
+      <DialogContent
+        className="p-0"
+        onOpenAutoFocus={(event) => {
```
EmilMalanczak, line 16: it seems to be annoying issue maybe we should add a prop with meaningful name for DialogContent component?

EmilMalanczak (approved):

packages/app/src/ui/organisms/asset-selector-with-input/AssetSelectorWithInput.tsx
```diff
@@ -40,12 +41,15 @@ export function AssetSelectorWithInput<TFieldValues extends FieldValues>({
   walletIconLabel,
   maxSelectedFieldName,
 }: AssetSelectorWithInputProps<TFieldValues>) {
+  if (selectorAssets.length > 1 && !setSelectedAsset) {
```
krzkaczor, line 44: use assert instead!

krzkaczor (approved):

packages/app/src/ui/organisms/asset-selector-with-input/AssetSelectorWithInput.tsx
oskarvu: I think this condition is easier to read. With assert, we would need to use `!(selectorAssets.length > 1 && !setSelectedAsset)` or `selectorAssets.length === 1 || setSelectedAsset`.

PR #263, opened 2024-09-12
Update navbar

EmilMalanczak:


oskarvu:
Can you make a small adjustments:

- Match style of dropdown items with designs (so we don't have this blue vertical bar next to active item).
Screenshot 2024-09-12 at 11 21 07

- Fix behavior of accordion, so it doesn't push logo up when it's opened. 

https://github.com/user-attachments/assets/fb04ea2d-62bf-44b7-b01c-a4e6535826b3



- Optional: Maybe we could align 'x' with chevron. It looks a bit weird, that they are not aligned
Screenshot 2024-09-12 at 11 29 50

EmilMalanczak:
this vertical bar is included on designs https://www.figma.com/design/mJ0IlSLdrYzDnmUAMWls5A/Spark-App?node-id=5774-2466&t=IoY84INtO1E77E3S-4

but good catch with the mobile view

krzkaczor (approved):

PR #272, opened 2024-09-13
Add chain sensitivity for dialogs

EmilMalanczak:


packages/app/src/domain/hooks/useChainSensitive.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface UseChainSensitiveParams {
+  onChainChange: (chain: number) => void
+  chainId: number
+}
+
+export function useChainSensitive({ onChainChange, chainId }: UseChainSensitiveParams): boolean {
+  const [frozenChainId] = useState(chainId)
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: frozenChainId will never change
+  useEffect(() => {
```
EmilMalanczak, line 12: https://github.com/wevm/wagmi/discussions/1806
possible another approach however this would only detect that the chain is changed - we don't care which one

on current one we can create utility component eg. `ChainSensitiveWrapper` that will work like conditional rendering *only* for the selected chain eg. when we would render specific component only for the mainnet

oskarvu: I wonder if we need such granularity in deciding which dialog should be open and which not. All the dialogs that are opened using dialog dispatcher are dependent on chain specific data, so we probably should close them on chain change.

I would go with a simple solution for now, just closing opened dialog on chain change, so we don't need to think too much about if given dialog is safe to render and if underlying data providers will work for switched chain.

```
const wagmiConfig = useConfig()
const closeDialog = useCloseDialog()
useEffect(function closeDialogOnChainChange() {
  const unwatch = watchChainId(wagmiConfig, {
    onChange: closeDialog,
  })
  return unwatch
}, [])
 ```

yivlad (approved):

oskarvu (approved):

PR #274, opened 2024-09-16
Move usds related contracts to wagmi config

yivlad:


packages/app/src/pages/Savings.PageObject.ts
```diff
@@ -128,9 +128,15 @@ export class SavingsPageObject extends BasePageObject {
   }: { sDaiBalance: string; estimatedDaiValue: string }): Promise<void> {
     await expect(this.locatePanelByHeader('Savings DAI').getByTestId(testIds.savings.sDaiBalance)).toHaveText(
       sDaiBalance,
+      {
+        timeout: 60_000, // potentially should wait a bit for balance to reach the value because of timing issues in e2e tests
```
oskarvu, line 132: nit: extract to const and reuse

packages/app/src/config/abis/susdsAbi.ts
```diff
@@ -0,0 +1,29 @@
+export const susdsAbi = [
```
krzkaczor, line 1: isn't it better to simply inline this into wagmi config?

packages/app/src/config/chain/index.ts
oskarvu, line 1: I think you need to add entries to mainnet config, like: `usds`, `susds` symbols, `savingsUsdsInfoQuery`, extra tokens. Generally what we included in devnet config.

packages/app/src/config/abis/susdsAbi.ts
yivlad: What do you mean by inlining into wagmi config?

krzkaczor: i meant generated output*. You removed it from there because wagmi doesnt read it from etherscan correctly, but we could simply copy it manually and then wagmi wouldn't query it again, right?

yivlad: But it will overwrite it on every wagmi cli run. I don't think it's a good solution

krzkaczor: mhm okay

oskarvu: I don't like idea of inlining anything into generated output.

packages/app/src/pages/Savings.PageObject.ts
krzkaczor: doesnt' matter. we will rewrite these tests to use mainnet and situation should improve

krzkaczor (approved):

packages/app/src/config/chain/index.ts
```diff
@@ -161,13 +173,9 @@ export function getChainConfigEntry(chainId: number): ChainConfigEntry {
...
     return {
       ...mainnetConfig,
-      USDSSymbol: TokenSymbol('USDS'),
-      sUSDSSymbol: TokenSymbol('sUSDS'),
-      savingsUsdsInfoQuery: mainnetSavingsUsdsInfoQuery,
       meta: getUSDSDevChainMeta(mainnetConfig.meta),
-      savingsInputTokens: [...mainnetConfig.savingsInputTokens, TokenSymbol('USDS')],
       extraTokens: [
-        ...mainnetConfig.extraTokens,
+        ...mainnetConfig.extraTokens.filter(({ symbol }) => !['USDS', 'sUSDS'].includes(symbol)),
```
EmilMalanczak, line 178: maybe we could use `mapMainnetAddressToDevnet` here?

EmilMalanczak (approved):

oskarvu (approved):

packages/app/src/test/e2e/injectSetup.ts
```diff
@@ -81,3 +86,30 @@ function overrideDateClass(fakeNow: number): void {
...
+  })
+  const susdsBytecode = await publicClient.getBytecode({ address: susdsAddress })
+
+  await page.addInitScript(
+    ({ PLAYWRIGHT_USDS_CONTRACTS_NOT_AVAILABLE_KEY, susdsBytecode }) => {
+      if (susdsBytecode === undefined || susdsBytecode.length <= 2) {
+        ;(window as any)[PLAYWRIGHT_USDS_CONTRACTS_NOT_AVAILABLE_KEY] = true
+      }
+    },
+    { PLAYWRIGHT_USDS_CONTRACTS_NOT_AVAILABLE_KEY, susdsBytecode },
```
krzkaczor, line 113: IMO extract  this to variables like `usdsAvaiaible`:
```
if (susdsBytecode === undefined || susdsBytecode.length <= 2) {
        ;(window as any)[PLAYWRIGHT_USDS_CONTRACTS_NOT_AVAILABLE_KEY] = true
      }
```

and then pass only this bool to `addInitScript`

packages/app/src/config/chain/index.ts
```diff
@@ -68,13 +74,13 @@ const chainConfig: ChainConfig = {
...
+    savingsUsdsInfoQuery: PLAYWRIGHT_USDS_CONTRACTS_NOT_AVAILABLE ? undefined : mainnetSavingsUsdsInfoQuery,
     daiSymbol: TokenSymbol('DAI'),
     sDaiSymbol: TokenSymbol('sDAI'),
-    USDSSymbol: undefined,
-    sUSDSSymbol: undefined,
+    USDSSymbol: PLAYWRIGHT_USDS_CONTRACTS_NOT_AVAILABLE ? undefined : TokenSymbol('USDS'),
+    sUSDSSymbol: PLAYWRIGHT_USDS_CONTRACTS_NOT_AVAILABLE ? undefined : TokenSymbol('sUSDS'),
     mergedDaiAndSDaiMarkets: true,
-    savingsInputTokens: [TokenSymbol('DAI'), TokenSymbol('USDC')],
+    savingsInputTokens: [TokenSymbol('DAI'), TokenSymbol('USDC'), TokenSymbol('USDS')],
```
oskarvu, line 83: Should we also assert usds contracts are available before adding usds to this array? I guess tests are passing, but to be consistent.

krzkaczor: in test setup? i think we don't need to do that as tests are pasing anyway and this would require hardcoding USDS address in yet another place :/

yivlad: Good catch

krzkaczor (approved):

oskarvu (approved):

EmilMalanczak (approved):

PR #281, opened 2024-09-18
Add useOracleInfo hook

EmilMalanczak:


packages/app/src/features/market-details/components/oracle-panel/components/YieldingFixedOraclePanel.tsx
```diff
@@ -51,21 +67,21 @@ export function YieldingFixedOraclePanel({ oracle, chainId }: YieldingFixedOracl
...
-              <BlockExplorerAddressLink address={zeroAddress} chainId={chainId} />
+              <BlockExplorerAddressLink address={token.address} chainId={chainId} />
             </InfoTile.Value>
           </InfoTile>
 
           <InfoTile>
             <InfoTile.Label>Token Contract</InfoTile.Label>
             <InfoTile.Value>
-              <BlockExplorerAddressLink address={zeroAddress} chainId={chainId} />
+              <BlockExplorerAddressLink address={baseToken.address} chainId={chainId} />
```
yivlad, line 77: IMO here we should display not the base token itself but the address of the price oracle for the base asset. I think what we want to display here is basically implementation of `latestAnswer` function. Here is the one from wstETH oracle:
```solidity
function latestAnswer() external view returns (int256) {
    int256 ethUsd       = ethSource.latestAnswer();
    int256 exchangeRate = int256(steth.getPooledEthByShares(1e18));

    if (ethUsd <= 0 || exchangeRate <= 0) {
       return 0;
    }

    return exchangeRate * ethUsd / 1e18;
}
```
Now we are lacking the part with `ethSource`.

packages/app/src/domain/market-info/marketInfo.ts
```diff
@@ -84,7 +85,7 @@ export interface Reserve {
   variableBorrowApy: Percentage | undefined
 
   priceInUSD: BigNumber
-
+  priceOracle: Address
```
yivlad, line 88:
```suggestion
  priceOracle: CheckedAddress
```

packages/app/src/domain/oracles/types.ts
```diff
@@ -0,0 +1,14 @@
+import { ReserveOracleType } from '@/config/chain/types'
+import { Address } from 'viem'
+import { NormalizedUnitNumber } from '../types/NumericValues'
+import { Token } from '../types/Token'
+
+export type OracleInfo = {
```
yivlad, line 6: I think union type with `baseToken` and `ratio` present only when oracle type is `yielding-fixed` would be safer.

packages/app/src/config/chain/index.ts
```diff
@@ -140,19 +143,57 @@ const chainConfig: ChainConfig = {
...
+        ratio: async ({ reserve, wagmiConfig }) => {
+          const ratio = await readContract(wagmiConfig, {
+            abi: weEthRatioAbi,
+            address: reserve.token.address,
+            functionName: 'getRate',
+            args: [],
+            chainId: mainnet.id,
+          })
+
+          return NormalizedUnitNumber(BaseUnitNumber(ratio).dividedBy(10 ** reserve.token.decimals))
```
yivlad, line 195: nit: 
```suggestion
          return NormalizedUnitNumber(BaseUnitNumber(ratio).shiftedBy(-reserve.token.decimals))
```
is more canonical in our codebase

packages/app/src/domain/oracles/useOracleInfo.ts
```diff
@@ -0,0 +1,33 @@
...
+  marketInfo: MarketInfo
+}
+export type UseOracleInfoResult = SuspenseQueryWith<{
+  oracleInfo: OracleInfo
+}>
+
+export function useOracleInfo({ reserve, marketInfo }: UseOracleInfoParams): UseOracleInfoResult {
+  const wagmiConfig = useConfig()
+
+  const res = useSuspenseQuery({
```
yivlad, line 21: I think we wanted this to be non-blocking, e.g. not using suspense

packages/app/src/config/chain/index.ts
krzkaczor: Isn't `token.fromBaseUnit` the way to go 😬

packages/app/src/config/chain/index.ts
```diff
@@ -140,19 +143,57 @@ const chainConfig: ChainConfig = {
         type: 'yielding-fixed',
```
krzkaczor, line 143: WETH (above) oracle is providedBy chainlink & chronicle (aggor)

packages/app/src/features/market-details/components/oracle-panel/components/YieldingFixedOraclePanel.tsx
krzkaczor: yes, especially that for eth it's currectly linked to 0xeeeee  (fake ETH address) :( 
image

packages/app/src/domain/oracles/query.ts
```diff
@@ -0,0 +1,52 @@
...
+        priceOracleAddress: reserve.priceOracle,
+        chainId: marketInfo.chainId,
+        baseTokenReserve: undefined,
+        ratio: undefined,
+      }
+
+      if (oracleConfig?.type === 'yielding-fixed') {
+        const baseTokenReserve = marketInfo.findReserveBySymbol(oracleConfig.baseAsset)
+
+        assert(baseTokenReserve, `Base token ${oracleConfig.baseAsset} not found`)
```
EmilMalanczak, line 35: I'm not sure if that is safe to use assert here - used it as a type guard

packages/app/src/domain/oracles/types.ts
EmilMalanczak: Tried to switch to union twice but wasn't satisfied with result, also I was struggling with typing because of union of unions.

packages/app/src/domain/oracles/query.ts
yivlad: The code can be replaced with:

```ts
const baseTokenReserve = marketInfo.findOneReserveBySymbol(oracleConfig.baseAsset)
```

I think it's alright if we make the assumption that base asset and derivatives use the same oracle under the hood. I'm not sure this is a good assumption though, as for instance for wstETH `ethSource` contract is not the one that is used by WETH reserve

krzkaczor: > I'm not sure this is a good assumption though, as for instance for wstETH ethSource contract is not the one that is used by WETH reserve

This is surprising 😮 I told @EmilMalanczak do it like this but you're right, this is not correct. I wonder why is that. I think it's fine to merge this like this as this PR is already getting crowded but we should sort this out.

packages/app/src/domain/oracles/oracleRatioFetchers.ts
```diff
@@ -0,0 +1,88 @@
...
+import { readContract } from 'wagmi/actions'
+import { Reserve } from '../market-info/marketInfo'
+import { BaseUnitNumber, NormalizedUnitNumber } from '../types/NumericValues'
+
+export interface OracleRatioFetcherParams {
+  reserve: Reserve
+  wagmiConfig: Config
+}
+
+export async function fetchWstEthRatioMainnet({
```
krzkaczor, line 21: let's treat token names as a single word so it should be: `fetchWstethRatioMainnet`

packages/app/src/domain/oracles/oracleRatioFetchers.ts
```diff
@@ -0,0 +1,88 @@
...
+    address: stethAddress,
+    functionName: 'getPooledEthByShares',
+    args: [toBigInt(10 ** reserve.token.decimals)],
+    chainId: mainnet.id,
+  })
+
+  return reserve.token.fromBaseUnit(BaseUnitNumber(ratio))
+}
+
+export async function fetchREthRatio({
```
krzkaczor, line 44: `fetchRethRatio` etc.

packages/app/src/domain/oracles/oracleRatioFetchers.ts
```diff
@@ -0,0 +1,88 @@
...
+export interface OracleRatioFetcherParams {
+  reserve: Reserve
+  wagmiConfig: Config
+}
+
+export async function fetchWstEthRatioMainnet({
+  reserve,
+  wagmiConfig,
+}: OracleRatioFetcherParams): Promise<NormalizedUnitNumber> {
+  const stethAddress = await readContract(wagmiConfig, {
```
krzkaczor, line 25: we are positive that this address will never change. We should cache this value "forver" with `react-query`. How to do this?

krzkaczor (approved):
Please add a feature flag

yivlad (approved):

oskarvu (approved):

PR #291, opened 2024-09-24
Display unsupported chain view at farms

oskarvu:
Closes FRO-987

While working on this PR, we noticed regression with sandbox modal. Dialog shouldn't be closed when chain is changing. Related [ticket](https://linear.app/phoenixlabs/issue/FRO-994/make-dialog-close-on-chain-change-configurable)

### DIFF
--- a/packages/app/src/domain/farms/farmsInfo.ts
+++ b/packages/app/src/domain/farms/farmsInfo.ts
@@ -3,10 +3,10 @@ import { CheckedAddress } from '../types/CheckedAddress'
 import { Farm } from './types'
 
 export class FarmsInfo {
-  hasFarms: boolean
+  constructor(private farms: Farm[]) {}
 
-  constructor(private farms: Farm[]) {
-    this.hasFarms = farms.length > 0
+  get hasFarms(): boolean {
+    return this.farms.length > 0
   }
 
   findFarmByAddress(address: CheckedAddress): Farm | undefined {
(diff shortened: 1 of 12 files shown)


yivlad (approved):

packages/app/src/domain/farms/farmsInfo.ts
```diff
@@ -5,6 +5,10 @@ import { Farm } from './types'
 export class FarmsInfo {
   constructor(private farms: Farm[]) {}
 
+  get hasFarms(): boolean {
```
krzkaczor, line 8: IMO I would avoid using getters and made it a simple function instead... There is no good reason to use getters here (and in most cases xd)

packages/app/src/domain/hooks/useUnsupportedChain.ts
```diff
@@ -0,0 +1,32 @@
...
+  isGuestMode: boolean
+  openConnectModal: () => void
+  openChainModal: () => void
+  openSandboxModal: () => void
+  switchChain: (chainId: number) => void
+}
+
+export function useUnsupportedChain(): UseUnsupportedChainResult {
+  const openDialog = useOpenDialog()
+  const isGuestMode = useAccount().isConnected === false
```
EmilMalanczak, line 16: I think that would be better to simply destructure this first

const { isConnected } = useAccount()
const isGuestMode = isConnected === false

EmilMalanczak (approved):

packages/app/src/domain/hooks/useUnsupportedChain.ts
oskarvu: Shouldn't matter here, especially that we don't use `isConnected` anywhere else in this file.

PR #293, opened 2024-09-24
Fix sandbox mode on gnosis

oskarvu:
Closes FRO-994


packages/app/src/features/dialogs/dispatcher/DialogDispatcherContainer.tsx
```diff
@@ -1,20 +1,22 @@
...
   const closeDialog = useCloseDialog()
 
-  const isChainMatching = useChainSensitive({
-    onChainChange: closeDialog,
-  })
-
-  if (!openedDialog || !isChainMatching) {
+  if (!openedDialog) {
     return null
   }
```
EmilMalanczak, line 12: previously I've done that in this way but we decided to switch from this approach into moving `useCloseDialogOnChainChange` equivalent into `DialogDispatcherContainer` body. Was there a problem with this?

oskarvu: Restructured code a bit.

EmilMalanczak (approved):

yivlad (approved):

PR #294, opened 2024-09-24
Tweak playwright configuration to fail after few failures

oskarvu:
It looks like fix is not yet available on latest release. I looked at [PR](https://github.com/microsoft/playwright/issues/26350), and it has label `1.48` attached, while the current release is `1.47.2`. I suspect it might take some time until available.

EDIT: I've patched playwright to include given fix.

### DIFF
--- a/packages/app/playwright.config.ts
+++ b/packages/app/playwright.config.ts
@@ -21,7 +21,7 @@ export default defineConfig({
     timeout: 20_000,
   },
   timeout: 60_000 * 3, // sometimes tenderly can be slow
-  maxFailures: 1, // should mark test as failed only after all retires are exhausted
+  maxFailures: 3, // should mark test as failed only after all retires are exhausted
 
   webServer: {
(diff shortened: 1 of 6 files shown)


packages/app/playwright.config.ts
```diff
@@ -21,7 +21,7 @@ export default defineConfig({
     timeout: 20_000,
   },
   timeout: 60_000 * 3, // sometimes tenderly can be slow
-  maxFailures: undefined, // don't use this as it doesn't respect retires
+  maxFailures: 1, // should mark test as failed only after all retires are exhausted
```
krzkaczor, line 24: IMO we can set it to something like 3 or even 5 so we get list of many failing tests not just the first one. I think it will aid the debugging.

krzkaczor (approved):

EmilMalanczak (approved):

yivlad (approved):

PR #301, opened 2024-09-25
Rename stake to deposit and unstake to withdraw

yivlad:


packages/app/src/features/farm-details/dialogs/stake/views/StakeView.tsx
```diff
@@ -41,7 +41,7 @@ export function StakeView({
 
   return (
     <MultiPanelDialog panelRef={ref}>
-      <DialogTitle>Stake in {farm.rewardToken.symbol} Farm</DialogTitle>
+      <DialogTitle>Deposit to {farm.rewardToken.symbol} Farm</DialogTitle>
```
EmilMalanczak, line 44: shouldn't be `into` as fixed in `StakeActionRow`

EmilMalanczak (approved):

oskarvu (approved):

packages/app/src/features/farm-details/dialogs/stake/views/StakeView.tsx
oskarvu: Most correct would be "Deposit to the farm", but whatever.

PR #302, opened 2024-09-25
Add max withdraw and exit from farms tests

oskarvu:
Closes FRO-984

### DIFF
--- a/packages/app/src/features/farm-details/dialogs/unstake/UnstakeDialog.PageObject.ts
+++ b/packages/app/src/features/farm-details/dialogs/unstake/UnstakeDialog.PageObject.ts
@@ -14,28 +14,86 @@ export class UnstakeDialogPageObject extends DialogPageObject {
       state: 'detached',
     })
   }
+
+  async clickExitFarmSwitchAction(): Promise<void> {
+    await this.page.getByTestId(testIds.farmDetails.unstakeDialog.exitFarmSwitchPanel.switch).click()
+  }
   //
 
   // #region assertions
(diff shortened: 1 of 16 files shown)


packages/app/src/features/farm-details/dialogs/unstake/UnstakeDialog.PageObject.ts
```diff
@@ -14,47 +14,126 @@ export class UnstakeDialogPageObject extends DialogPageObject {
...
     await expect(this.page.getByTestId(testIds.farmDetails.unstakeDialog.transactionOverview.outcome)).toContainText(
-      transactionOverview.outcome,
+      outcome,
     )
   }
 
+  async expectExitTransactionOverview({ route, reward, outcome }: ExitTransactionOverview): Promise<void> {
+    await this.expectRoutesOverview(route)
+
+    const regexp = /([\d,\.]+) (\w+) \(\$([\d,\.]+)\) \+ ~([\d,\.]+) (\w+) \(~\$([\d,\.]+)\)/
```
oskarvu, line 80: Reward values are not precise in test, which already was a problem when claim tests were written. Thus, for outcome containing reward value, I needed to each value and assert that reward is in a range.

At first I added separate test ids to each value in component, but then it become super ugly and it was impossible to see what's going on with a text content (all the spans on each value to just add testid). So I went with regexp in the end.

yivlad (approved):

EmilMalanczak (approved):

PR #303, opened 2024-09-25
Restructure active farm info panel and add sandbox redirects

yivlad:


packages/app/src/features/farm-details/logic/useSandboxFarmRedirect.ts
```diff
@@ -0,0 +1,29 @@
...
+import { watchChainId } from 'wagmi/actions'
+import { useFarmDetailsParams } from './useFarmDetailsParams'
+
+export function useSandboxFarmRedirect(): void {
+  const config = useConfig()
+  const { address: farmAddress, chainId: farmChainId } = useFarmDetailsParams()
+  const { sandboxChainId, originChainId } = useSandboxState()
+  const navigate = useNavigate()
+
+  useEffect(() => {
```
EmilMalanczak, line 15: maybe we could reuse `useChainSensitive` hook here?

yivlad: `watchChainId` has convenient API with new `chainId` and `prevChainId`, so decided to stick to it

EmilMalanczak: so it does with `frozenChainId` as equiv to `prevChainId `

yivlad: Yeah, probably you're right, implementation of `useChainSensitive` can be adjusted to fit the needs here. Maybe we should make consistent, but let's do it later

oskarvu:
I think it will be better to format TVL as compact, so it fits nicer.

oskarvu (approved):

packages/app/src/features/farm-details/logic/useSandboxFarmRedirect.ts
```diff
@@ -0,0 +1,29 @@
+import { paths } from '@/config/paths'
+import { useSandboxState } from '@/domain/sandbox/useSandboxState'
+import { useEffect } from 'react'
+import { generatePath, useNavigate } from 'react-router-dom'
+import { useConfig } from 'wagmi'
+import { watchChainId } from 'wagmi/actions'
+import { useFarmDetailsParams } from './useFarmDetailsParams'
+
+export function useSandboxFarmRedirect(): void {
```
krzkaczor, line 9: i don't like that very much... this should be a generic mechanism so entering to sandbox mode on any page works. (what about market details page?). This should be improved in the future.

packages/app/src/features/farm-details/dialogs/unstake/logic/useUnstakeDialog.ts
```diff
@@ -44,7 +44,10 @@ export interface ExitFarmSwitchInfo {
 }
 
 export function useUnstakeDialog({ farm, initialToken }: UseStakeDialogParams): UseUnstakeDialogResult {
-  const { timestamp } = useTimestamp()
+  const { timestamp, updateTimestamp } = useTimestamp()
```
oskarvu, line 47: Probably worth to extract to separate hook in future. `useNowTimestamp` or something like this.

EmilMalanczak (approved):

oskarvu (approved):

PR #304, opened 2024-09-26
Clear sandbox when expired

oskarvu:
Closes FRO-997


packages/app/src/domain/sandbox/useUpToDateSanboxNetwork.ts
```diff
@@ -0,0 +1,34 @@
+import { tryOrDefault } from '@/utils/tryOrDefault'
+import { useStore } from '../state'
+import { SandboxNetwork } from '../state/sandbox'
+
+const SANDBOX_EXPIRY = 6 * 3600 * 1000 // 6 hours
+
+export function useUpToDateSandboxNetwork(): SandboxNetwork | undefined {
```
EmilMalanczak, line 7: why not adjust `unPersistSandboxSlice` logic?

oskarvu: Good point 👍🏻

packages/app/src/domain/state/sandbox.ts
```diff
@@ -61,6 +63,14 @@ export function unPersistSandboxSlice(persistedState: DeepPartial<PersistedSandb
     return {}
   }
 
+  if (isPersistedSandboxExpired(persistedState)) {
+    return {
```
EmilMalanczak, line 67: I think we can just return empty object as above or even put that into the same if

EmilMalanczak (approved):

yivlad (approved):

PR #314, opened 2024-09-27
Fix reverting to snapshots in E2E tests

yivlad:


oskarvu (approved):

packages/app/src/domain/sandbox/request.ts
```diff
@@ -1,16 +1,16 @@
...
-): Promise<void>
-export async function request(forkUrl: string, method: 'evm_snapshot', params: []): Promise<string>
-export async function request(forkUrl: string, method: 'evm_revert', params: [string]): Promise<void>
-export async function request(forkUrl: string, method: 'evm_increaseTime', params: [number]): Promise<void>
-export async function request(forkUrl: string, method: 'evm_setNextBlockTimestamp', params: [number]): Promise<void>
+): Promise<any>
+export async function request(forkUrl: string, method: 'evm_snapshot', params: []): Promise<any>
+export async function request(forkUrl: string, method: 'evm_revert', params: [string]): Promise<any>
+export async function request(forkUrl: string, method: 'evm_increaseTime', params: [number]): Promise<any>
+export async function request(forkUrl: string, method: 'evm_setNextBlockTimestamp', params: [number]): Promise<any>
```
EmilMalanczak, line 13: i think that would be better to use `unknown` instead

krzkaczor: true, but i guess we won't fix this in this PR. btw. i think we should refactor this to custom viem client in the future.

krzkaczor (approved):

packages/app/src/test/e2e/forking/setupFork.ts
```diff
@@ -106,6 +110,14 @@ export function setupFork(options: SetupForkOptions): ForkContext {
 
   test.beforeEach(async () => {
     await tenderlyRpcActions.revertToSnapshot(forkContext.forkUrl, forkContext.initialSnapshotId)
+    if (forkContext.isVnet) {
+      const client = createPublicClient({
+        chain: mainnet, // @todo select chain based on chainId
+        transport: http(forkContext.forkUrl),
+      })
+      const block = await client.getBlock()
```
krzkaczor, line 118: i think this deserves a comment explaining why it's not -1 block. At at least that it's not a mistake 😅 alternatively, we could save loaded block.timestamp together with snapshot id....

krzkaczor (approved):

EmilMalanczak (approved):

PR #319, opened 2024-10-01
Add MyEarningsChart component

EmilMalanczak:


packages/app/src/features/savings/components/savings-graph/utils.ts
oskarvu, line 1: It's the same util as in farm details Chart component. We should extract one and reuse it in both places.

packages/app/src/features/savings/components/savings-graph/components/MyEarningsGraph.tsx
```diff
@@ -0,0 +1,281 @@
...
+function TooltipContent({ data }: { data: GraphDataPoint }) {
+  const isPrediction = isDataPointPrediction(data)
+
+  return (
+    <div className="flex flex-col gap-3 rounded-xl border border-slate-700/10 bg-white p-3 shadow">
+      <div className="flex flex-col gap-3 text-slate-500 text-xs leading-none">{formatTooltipDate(data.date)}</div>
+      <div className="flex items-center gap-1.5 text-sm leading-none">
+        <Circle size={8} fill={isPrediction ? colors.secondary : colors.primary} stroke="0" />
+        <div>
+          {isPrediction && 'Future '}Savings:{' '}
```
oskarvu, line 248: Maybe we should use different wording here, like `Savings Prediction`? Using `Future Savings` maybe be misleading, since we cannot guarantee that value.

packages/app/src/features/savings/components/savings-graph/components/MyEarningsGraph.tsx
oskarvu, line 1: Consider renaming to `MyEaningsChart`.

packages/app/src/features/savings/components/savings-chart/utils.ts
yivlad, line 1: This one can be extracted and reused

yivlad (approved):

oskarvu (approved):

krzkaczor (approved):

PR #329, opened 2024-10-03
Add savings chart info hook

EmilMalanczak:


packages/app/src/domain/savings-charts/useSavingsRateInfo/useSavingsRateInfo.ts
```diff
@@ -0,0 +1,62 @@
...
+}: UseSavingsRateInfoParams): UseSavingsRateInfoResult {
+  const savingsRateInfo = useQuery({
+    ...savingsRateQueryOptions({
+      chainId,
+    }),
+  })
+
+  const savingsRateInfoFiltered = useQuery({
+    queryKey: [...savingsRateInfoQueryKey({ chainId }), timeframe, currentTimestamp],
+    queryFn: () => {
```
EmilMalanczak, line 32: as I look right now - will move this into `savingsRateFilteredQueryOptions`

in general - this is using cached response for `savingsRate` but to prevent filtering with each rerender I've put into its own query key - by this we reduce amount of filtering with each timeframe change within selected timestamp (1h because data is updating once a day)

EmilMalanczak: also noticed lack of `staleTime` for both hooks

packages/app/src/domain/savings-charts/useMyEarningsInfo/query.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export function myEarningsInfoQueryKey({ chainId, address }: Omit<MyEarningsQueryParams, 'staleTime'>): unknown[] {
+  return ['my-earnings', chainId, address]
+}
+
+const myEarningsDataResponseSchema = z
+  .array(
+    z.object({
+      // balance is sum of sdai_balance and susds_balance but since we display it separately we can use balance
```
oskarvu, line 48:
```suggestion
      // @note: response balance is a sum of sdai and usds balance, but we display chart only when user has only one token. Thus, we can treat this sum as a single token balance.
```

packages/app/src/domain/savings-charts/useMyEarningsInfo/query.ts
```diff
@@ -0,0 +1,113 @@
...
+  currentTimestamp,
+  timeframe,
+  address,
+  myEarningsInfo,
+  savingsInfo,
+}: MyEarningsFilteredQueryParams) {
+  return queryOptions({
+    queryKey: [...myEarningsInfoQueryKey({ chainId, address }), timeframe, currentTimestamp],
+    queryFn: () => {
+      assert(myEarningsInfo, 'My earnings info should be loaded before filtering')
```
oskarvu, line 85: skipToken

packages/app/src/domain/savings-charts/useMyEarningsInfo/query.ts
```diff
@@ -0,0 +1,113 @@
...
+  staleTime: number
+  currentTimestamp: number
+  timeframe: Timeframe
+  myEarningsInfo?: MyEarningsInfoItem[]
+  address?: Address
+  savingsInfo: SavingsInfo
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export function myEarningsFilteredQueryOptions({
```
yivlad, line 73: Why do we need to use react-query for filtering?

packages/app/src/domain/savings-charts/useMyEarningsInfo/calculate-predictions.ts
```diff
@@ -0,0 +1,83 @@
...
+import { NormalizedUnitNumber } from '@/domain/types/NumericValues'
+import { Timeframe } from '@/ui/charts/defaults'
+import { range } from '@/utils/array'
+import { assertNever } from '@/utils/assertNever'
+import { MyEarningsInfoItem } from './types'
+
+const SECONDS_PER_DAY = 24 * 60 * 60
+
+export interface CalculatePredictionsParams {
+  timestamp: number // in ms
```
yivlad, line 11: We use pattern with naming `timestampInMs`

packages/app/src/domain/savings-charts/useMyEarningsInfo/calculate-predictions.ts
```diff
@@ -0,0 +1,83 @@
...
+    savingsInfo.predictSharesValue({ timestamp: timestamp + SECONDS_PER_DAY, shares: balance }).minus(base),
+  )
+
+  const optimalPredictionsLength = Math.ceil(dataLength * 0.33)
+
+  switch (timeframe) {
+    case '7D':
+      return calculatePredictionsIncomeByDays({
+        // Ensure to have proportional length of projections to the data length
+        days: dataLength > 7 ? 3 : optimalPredictionsLength,
```
yivlad, line 37: Isn't it basically the same as `Math.max(optimalPredictionsLength, 3)`? And I think it's much clearer

packages/app/src/domain/savings-charts/useSavingsChartsInfoQuery.ts
```diff
@@ -0,0 +1,53 @@
...
+  const { address } = useAccount()
+  const { timestamp } = useTimestamp({ refreshIntervalInMs: STEP_IN_MS })
+
+  const myEarningsInfo = useMyEarningsInfo({
+    address,
+    chainId,
+    timeframe: selectedTimeframe,
+    currentTimestamp: timestamp,
+    savingsInfo,
+    staleTime: STEP_IN_MS,
```
yivlad, line 37: Why do we need it?

packages/app/src/domain/savings-charts/useSavingsRateInfo/query.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export function savingsRateInfoQueryKey({ chainId }: Omit<SavingsRateQueryParams, 'staleTime'>): unknown[] {
+  return ['savings-rate', chainId]
+}
+
+const savingsRateDataResponseSchema = z
+  .array(
+    z.object({
+      date: z.string().transform((value) => new Date(value)),
+      rate: z.string().transform((value) => NormalizedUnitNumber(value)),
```
oskarvu, line 43: Take a look at `packages/app/src/domain/common/validation.ts` and try to use it in your code. You can also add some validation there. For example this `rate: z.string().transform((value) => NormalizedUnitNumber(value))` could be just `rate: normalizedUnitNumberSchema`.

packages/app/src/domain/savings-charts/useMyEarningsInfo/query.ts
EmilMalanczak: left a comment earlier but resolved it:

in general - didn't want to add `timeframe` dependency to the `myEarningsInfo` and `savingsRateInfo`

The data returned eg. `savingsRate` is containing data since 2019 so it's pretty large set of data. By passing filtering into the `select` function I would be forced to add `timeframe` to dependency. 

According to my research - `select` runs when: 
- query is first fetched
- query data is updated
- data is invalidated and refetched
 
 so to achieve update with each timeframe change we would have to download full data and put it into cache.

To prevent this I created separate query that takes the cached data so we download the data only once and by storing it into cache we reduce amount of filtering within selected timestamp (1h because data is updating once a day)

packages/app/src/domain/savings-charts/useSavingsChartsInfoQuery.ts
EmilMalanczak: This data is only updated once per day, so we can extend the `staleTime` beyond the default (which is like 1-2 minutes). Since we're dealing with large datasets that aren't updated frequently, there's no need to fetch them so often. Increasing the staleTime will help reduce unnecessary network requests and improve performance.

packages/app/src/domain/savings-charts/useMyEarningsInfo/query.ts
yivlad: Select function also runs each time a newly created function instance is passed to `useQuery`, so for instance:
```ts
useQuery({
  ...,
  select: (data) => data
})
```
here select function will be run on every render. You can memoize select with useCallback but pass timeframe to dependency array.
```ts
useQuery({
  ...,
  select: useCallback((data) => filter(data, timeframe), [timeframe])
})
```

But anyway I think even simple useMemo would have worked.

packages/app/src/domain/savings-charts/useMyEarningsInfo/calculate-predictions.ts
EmilMalanczak: actually that should be `in seconds` wrong comment

packages/app/src/domain/savings-charts/useMyEarningsInfo/calculate-predictions.ts
```diff
@@ -0,0 +1,83 @@
...
+export function calculatePredictions({
+  timestamp,
+  balance,
+  savingsInfo,
+  timeframe,
+  dataLength,
+}: CalculatePredictionsParams): MyEarningsInfoItem[] {
+  const base = savingsInfo.convertToAssets({ shares: balance })
+
+  const dayIncomePrediction = NormalizedUnitNumber(
```
krzkaczor, line 27: this can't be estimated this way since interests compounds

krzkaczor (changes requested):

krzkaczor (approved):

packages/app/src/utils/schemas.ts
```diff
@@ -0,0 +1,3 @@
+import * as z from 'zod'
+
+export const dateSchema = z.string().transform((value) => new Date(value))
```
yivlad, line 3:
```suggestion
export const dateSchema = z.coerce.date()
```

yivlad (approved):

PR #331, opened 2024-10-04
Assemble savings charts

EmilMalanczak:


packages/app/src/features/savings/components/savings-charts/UsdsSavingsCharts.tsx
```diff
@@ -0,0 +1,52 @@
...
+  savingsRateInfo: UseSavingsRateInfoResult
+}
+
+export function UsdsSavingsCharts({
+  selectedTimeframe,
+  setSelectedTimeframe,
+  myEarningsInfo,
+  savingsRateInfo,
+}: UsdsSavingsChartsProps) {
+  const shouldDisplayMyEarnings =
```
yivlad, line 21: Can we extract his to the top-level hook? Also I think we should just that there is at least one non-zero element

yivlad (approved):

packages/app/src/domain/savings-charts/useMyEarningsInfo/calculatePredictions.ts
```diff
@@ -9,44 +9,44 @@ const SECONDS_PER_DAY = 24 * 60 * 60
...
         timestamp,
         savingsInfo,
       })
 
     case '1Y':
     case 'All':
       return calculatePredictionsIncomeByDays({
-        days: Math.min(optimalPredictionsLength, 60),
-        balance,
+        // setting upper bounds due to visible performance problems
```
krzkaczor, line 47: are we recalculating thousands of points multiple times per second? 😬  we really need to prioritze fixing this ASAP.

packages/app/src/domain/savings-charts/useSavingsRateInfo/query.ts
```diff
@@ -37,12 +37,11 @@ const savingsRateDataResponseSchema = z
...
     z.object({
       date: dateSchema,
-      rate: normalizedUnitNumberSchema,
-      dsr_rate: normalizedUnitNumberSchema,
+      dsr_rate: percentageSchema,
       ssr_rate: z
         .string()
         .nullable()
-        .transform((value) => (value ? NormalizedUnitNumber(value) : null)),
+        .transform((value) => (value ? Percentage(value, true) : null)),
```
krzkaczor, line 44: According to used validators `ssr_rate` can be above 1 but `dsr_rate` no. Why?

packages/app/src/domain/savings-charts/useMyEarningsInfo/calculatePredictions.ts
EmilMalanczak: its memoized - performance meant in the comment is related to overhead while calculating large amount of predictions at once

EmilMalanczak:
All graph stories got affected by non-type safety on stories. 
I think we should update the guideline, current guide preserves only type safety for the values but it doesn't ensure that we pass all required props 

image

image

krzkaczor (approved):

PR #333, opened 2024-10-07
Add linear charts 

EmilMalanczak:


packages/app/src/domain/savings-charts/useMyEarningsInfo/calculatePredictions.ts
```diff
@@ -45,7 +45,7 @@ export function calculatePredictions({
     case 'All':
       return calculatePredictionsIncomeByDays({
         // setting upper bounds due to visible performance problems
-        days: Math.max(Math.min(optimalPredictionsLength, 360), 180),
+        days: Math.max(Math.min(optimalPredictionsLength, 270), 90),
```
EmilMalanczak, line 48: reduced the amount of calculations - maybe for timeframe `1Y / All ` we should consider calculating every 3rd day? `range` function already supports that

yivlad (approved):

oskarvu (approved):

PR #343, opened 2024-10-11
Simplify overlay handling and fix scroll bug

oskarvu:
Closes FRO-1051


oskarvu:
Possible to simplify even more with pure css.

PR #355, opened 2024-10-16
Handle navbar links active state

oskarvu:
Closes FRO-1048


packages/app/src/config/paths.ts
```diff
@@ -4,8 +4,8 @@ export const paths = {
   markets: '/markets',
   savings: '/savings',
   farms: '/farms',
-  marketDetails: '/market-details/:chainId/:asset',
-  farmDetails: '/farm-details/:chainId/:address',
+  marketDetails: '/markets/market-details/:chainId/:asset',
```
yivlad, line 7: How about switching to `'/markets/:chainId/:asset'` ? I think this is generally considered to be good naming practice. Also we'll get easier matching without need to add `/*` to the end of paths

packages/app/src/config/paths.ts
```diff
@@ -6,8 +6,8 @@ export const paths = {
   markets: '/markets',
   savings: '/savings',
   farms: '/farms',
-  marketDetails: '/market-details/:chainId/:asset',
-  farmDetails: '/farm-details/:chainId/:address',
+  marketDetails: '/markets/:chainId/:asset',
```
krzkaczor, line 9: This is a breaking change, making old links obsolete. We should always have redirects in place.

oskarvu: We have redirects in this PR.

yivlad (approved):

PR #361, opened 2024-10-17
Fix switching to sandbox in select network dialog

yivlad:


oskarvu:
There is an issue with dev mode sandbox, that I noticed testing this PR. Switching from production network to dev sandbox isn't working.

Steps to reproduce:
- connect wallet
- create dev sandbox (will switch network correctly to dev sandbox after creation)
- switch network to production one
- try switching back to dev sandbox

Works ok with normal sandbox.

yivlad:
@oskarvu good catch. Removing dev sandbox for now as it's rarely used anyways.

oskarvu (approved):

EmilMalanczak (approved):

PR #366, opened 2024-10-18
Add stables convert form UI + logic

yivlad:


packages/app/src/config/chain/types.ts
```diff
@@ -84,6 +84,7 @@ export interface ChainConfigEntry {
   sdaiSymbol: TokenSymbol | undefined
   usdsSymbol: TokenSymbol | undefined
   susdsSymbol: TokenSymbol | undefined
+  psmStables?: TokenSymbol[]
```
oskarvu, line 87: I think we should make prop required, so we explicitly see that it's not supported when looking at config.

packages/app/src/features/dialogs/convert-stables/components/form/ConvertStablesForm.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export function ConvertStablesForm({ formFields, form }: ConvertStablesFormProps) {
+  const {
+    selectedAsset1,
+    selectedAsset2,
+    changeAsset1,
+    changeAsset2,
+    asset1Options,
+    asset2Options,
```
oskarvu, line 20: nit:

```suggestion
    selectedFrom,
    selectedTo,
    changeFrom,
    changeTo,
    fromOptions,
    toOptions,
```
but please consider, cause it reads much better ;)

oskarvu (approved):

PR #373, opened 2024-10-22
Make connection to BA API only when available

EmilMalanczak:
Wasn't satisfied with putting urls for savings/charts into separate fields inside config - moved them into new `apiUrls` config to treat it a source of truth for all API urls per chain. 

I wasn't satisfied with how savings and charts URLs were in separate fields in the config, so I moved them into a new `apiUrls` section. Now, all API URLs per chain are in one place, making `apiUrls` the go-to spot for managing them. I find this as easier and more consistent approach


oskarvu:
I'm not sure if decoupling savings and farms api urls from their config sections is a good idea. We introduced this separation on purpose. Having farm config and looking for api url in there wouldn't be confusing at all in my opinion. What do you think?

EmilMalanczak:
It's still external API so in my mind it's more like a separate thing - don't mind putting that into farms/savings config, in case of any kind of changes related to API (eg. stop using BA) urls we will have to dig through the config.

packages/app/src/config/chain/types.ts
```diff
@@ -73,9 +76,15 @@ export interface MarketsConfig {
 
 export interface SavingsConfig {
   inputTokens: TokenSymbol[]
-  chartsSupported: boolean
   savingsDaiInfoQuery: SavingsInfoQuery | undefined
   savingsUsdsInfoQuery: SavingsInfoQuery | undefined
+  getSavingsRateApiUrl: GetApiUrl<void>
```
oskarvu, line 81: Can be just `savingsRateApiUrl: string | undefined`

packages/app/src/domain/savings-charts/useSavingsChartsInfoQuery.ts
```diff
@@ -38,11 +39,14 @@ export function useSavingsChartsInfoQuery({
...
-    getChainConfigEntry(chainId).savings ?? raise('Savings config is not defined on this chain')
+  const { savings } = getChainConfigEntry(chainId)
+
+  assert(savings, 'Savings are not supported on this chain')
+
+  const { getEarningsApiUrl, savingsRateApiUrl } = savings
 
   const myEarningsInfo = useMyEarningsInfo({
-    address,
+    address: address ? CheckedAddress(address) : address,
```
oskarvu, line 49: I know it means same thing, but to make it more explicit:
`address ? CheckedAddress(address) : undefined`

yivlad (approved):

oskarvu (approved):

PR #374, opened 2024-10-22
Increase NotSupportedWarning height according to design

EmilMalanczak:
I've put `96px` equiv. height - prev. height was equal to `78px`. Also `96px` is a target height from figma


oskarvu:
I would make it a bit higher, even if it'll be different from designs. Maybe `h-28` or even `h-32`. I think `96px` is still a bit to small to be easily noticed. I would go with `h-28` personally.

EmilMalanczak:
`h-28` works well too but imo its max we should go with. On `h-32` paddings starts to be bigger than content inside, that looks unnatural

oskarvu (approved):

yivlad (approved):

PR #378, opened 2024-10-22
Add gas limit increase for specific transaction types mechanism

yivlad:


packages/app/src/domain/hooks/useIncreasedGasLimit.ts
```diff
@@ -0,0 +1,130 @@
...
+        config.address &&
+        isAddressEqual(config.address, address) &&
+        config.functionName === functionName,
+    )
+
+  const { data, isLoading } = useQuery({
+    queryKey: ['estimate-gas', chainId, address, account, functionName, JSONStringifyRich(args)],
+    queryFn: async () => {
+      if (!client || !account || !abi || !functionName || !address) {
+        throw new Error('Client, account, abi, function name and address are required')
```
oskarvu, line 80: Do we care about this error or handle it at all? Maybe we could simplify a bit and use a skip token instead. This would remove throwing this error and using enable option for useQuery.

If we want to throw, than maybe use assertion instead.
```assert(client && account && abi && functionName && address, 'Client, account, abi, function name and address are required')```

oskarvu (approved):

EmilMalanczak (approved):

PR #387, opened 2024-10-27
Add new Button component

EmilMalanczak:
- changed `text-*` classes into `typography-*`. Reason: `twMerge` inside `cn` helper is confused with this naming while using text color class eg. `text-reskin-base-white` and filters out the second one. 

- updated colors inside tailwind config based on updates on figma

- button component, slight icons color outsync to figma but i think that should be resolved at the time of working with icons


packages/app/tailwind.config.ts
```diff
@@ -146,46 +156,98 @@ export default {
...
+                secondary: 'rgb(var(--error-600))',
+              },
+              info: {
+                primary: 'rgb(var(--magenta-500))',
+                secondary: 'rgb(var(--magenta-600))',
+              },
             },
           },
+          alpha: {
+            dialog: 'color-mix(in srgb, rgb(var(--base-black)) 40%, transparent)',
```
EmilMalanczak, line 222: `rgba(var(--base-black), 0.4)` doesnt work due to lack of `,`

packages/app/src/ui/atoms/button/NewButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import { cn } from '@/ui/utils/style'
+import { Loader } from '../loader/Loader'
+
+const buttonVariants = cva(
+  'relative isolate inline-flex select-none items-center justify-center gap-1 overflow-hidden whitespace-nowrap rounded-reskin-sm transition-colors focus-visible:bg-reskin-base-white focus-visible:text-reskin-neutral-950 focus-visible:outline-none focus-visible:ring focus-visible:ring-reskin-primary-200 focus-visible:ring-offset-0',
+  {
+    variants: {
+      variant: {
+        primary:
```
oskarvu, line 13: Can you use `cn` helper to break lines into smaller chunks, so it will be easier to read? Something like this:


```
cn(
  'before:-z-10 bg-gradient-spark-primary text-reskin-base-white',
  'before:absolute before:inset-0 active:before:bg-reskin-neutral-800',
  'focus-visible:before:bg-reskin-base-white hover:before:bg-reskin-neutral-950',
  'active:text-reskin-base-white before:transition-colors before:content-[""]',
),
```

oskarvu:
Can we add missing variants to this PR, to make it complete?

I think we're missing states: `focused` and `pressed`, and also `both icons`. At least in stories. Let's make stories complete.

packages/app/src/ui/atoms/button/NewButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+          'before:-z-10 bg-gradient-spark-primary text-reskin-base-white before:absolute before:inset-0 active:before:bg-reskin-neutral-800 focus-visible:before:bg-reskin-base-white hover:before:bg-reskin-neutral-950 active:text-reskin-base-white before:transition-colors before:content-[""]',
```
yivlad, line 14: If `before:content-[""]` really needed?

yivlad (approved):

packages/app/src/ui/atoms/button/NewButton.tsx
EmilMalanczak: `content: ""` is required to render pseudo element in the browser but didn't notice tailwind inject that with `before:` selector

packages/app/src/ui/atoms/button/NewButton.tsx
```diff
@@ -44,12 +44,24 @@ const buttonVariants = cva(
...
+  variants: {
+    size: {
+      l: 'icon-sm',
+      m: 'icon-sm',
+      s: 'icon-xs',
+    },
+  },
+})
+
+type IconType = React.FC<{ className?: string }>
```
oskarvu, line 57: Let's use `React.ComponentType<{ className? string }>`.

PR #388, opened 2024-10-28
Add select component

yivlad:


packages/app/src/ui/atoms/new/select/Select.tsx
```diff
@@ -0,0 +1,93 @@
+import * as SelectPrimitive from '@radix-ui/react-select'
+import * as React from 'react'
+
+import { assets } from '@/ui/assets'
+import { cn } from '@/ui/utils/style'
+
```
EmilMalanczak, line 6: maybe leave a comment that its not complete list of elements from shadcn in case we would need eg. Label element.

EmilMalanczak (approved):

oskarvu (approved):

PR #395, opened 2024-10-29
Add AssetInput component

yivlad:


packages/app/src/ui/organisms/new/asset-input/AssetInput.tsx
```diff
@@ -0,0 +1,162 @@
...
+  setSelectedAsset,
+  showError = true,
+  disabled = false,
+  maxValue,
+  maxSelectedFieldName,
+}: AssetInputProps<TFieldValues>) {
+  const { setValue, trigger, getValues } = useFormContext()
+
+  return (
+    <Controller
```
EmilMalanczak, line 38: What do you think about changing this into `useController`? It's same logic as `Controller` component but I think it's more clear to use in this way  [docs](https://www.react-hook-form.com/api/usecontroller/)

oskarvu (approved):

EmilMalanczak (approved):

PR #397, opened 2024-10-29
Update switch styles

oskarvu:
Closes FRO-1115

### DIFF
--- a/packages/app/tailwind.config.ts
+++ b/packages/app/tailwind.config.ts
@@ -41,6 +41,36 @@ export default {
         warning: 'rgb(var(--warning-600))',
         error: 'rgb(var(--error-700))',
       },
+      outlineColor: {
+        primary: 'rgb(var(--neutral-100))',
+        secondary: 'rgb(var(--neutral-200))',
+        tertiary: 'rgb(var(--neutral-300))',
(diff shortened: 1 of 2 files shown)


packages/app/tailwind.config.ts
```diff
@@ -41,6 +41,36 @@ export default {
...
+        secondary: 'rgb(var(--neutral-200))',
+        tertiary: 'rgb(var(--neutral-300))',
+        quaternary: 'rgb(var(--neutral-400))',
+        brand: {
+          primary: 'rgb(var(--primary-400))',
+          secondary: 'rgb(var(--primary-500))',
+          tertiary: 'rgb(var(--primary-600))',
+        },
+        system: {
+          success: {
```
EmilMalanczak, line 55: currently we have DEFAULT option only for `bg` and `fg` colors and for each palette. I think we should be consistent here

EmilMalanczak (approved):

packages/app/tailwind.config.ts
EmilMalanczak: actually ignore this - I will move this comment into #396

yivlad (approved):

PR #398, opened 2024-10-30
Add new health factor indicator

yivlad:


oskarvu:
Nice. There is only one little problem with how text under badge is rendered on mobile.
Screenshot 2024-10-31 at 11 11 02

oskarvu (approved):

PR #413, opened 2024-11-04
adjust old nav components

EmilMalanczak:


packages/app/.storybook/styles.css
yivlad, line 1: How is this different from decorator?

EmilMalanczak: that decorator didn't apply styles because of the dynamic class (the class was attached but tailwind didn't generate style to corresponding class, don't know the reason, neither why it worked before) - then noticed that when using `WithClassname` on the story level the style was applied but div underneath (created with decorator) didn't have min-height which was causing chromatic issue again. Thats why apply these styles with `> *`. Also I think it's more correct to use .css file for this case

packages/app/src/features/topbar/components/topbar-airdrop/TopbarAirdrop.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+      tokenReward: NormalizedUnitNumber('8.73949580999E-7'),
+      tokenRatePerSecond: NormalizedUnitNumber('2.8442E-13'),
+      tokenRatePrecision: 14,
+      refreshIntervalInMs: 100,
+      timestampInMs,
+    },
+  },
+}
+
+export const NotGrowing = {
```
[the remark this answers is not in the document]
EmilMalanczak: didn't check that - but yes. Storybook has overriden timestamp so useTimestamp returns the same timestamp over and over again. Removing this story

packages/app/src/domain/common/format.ts
```diff
@@ -59,3 +60,20 @@ export function findSignificantPrecision(
   const unitPriceUsd = Number(_unitPriceUsd)
   return Math.max(Math.floor(Math.log10(unitPriceUsd)) + desiredPrecisionOfUsd, 0)
 }
+
+export interface FormatAirdropAmountParams {
```
oskarvu, line 64: I wouldn't put it in domain and colocate with aridrop badge, since it's only used there.

oskarvu (approved):

PR #414, opened 2024-11-04
Use new inputs in dialogs

yivlad:


oskarvu (approved):

packages/app/src/features/dialogs/common/components/form/DialogForm.tsx
```diff
@@ -17,23 +16,21 @@ export interface DialogFormProps {
...
-      <DialogPanelTitle>Amount</DialogPanelTitle>
-      <AssetSelectorWithInput
+      <AssetInput
+        label="Amount"
         fieldName="value"
         control={form.control}
         selectorAssets={selectorAssets}
         selectedAsset={selectedAsset}
-        setSelectedAsset={changeAsset}
+        setSelectedAsset={changeAsset ?? (() => {})}
```
EmilMalanczak, line 30: better assign when destructuring eg. `{ changeAsset = () => {}}` - wouldn't that be better to allow this param to be optional?

EmilMalanczak (approved):

PR #415, opened 2024-11-04
assemble Topbar

EmilMalanczak:


yivlad (approved):

oskarvu:
Will you add airdop/rewards badges variants (displaying only icon on narrower views) in this PR or in a separate one?

oskarvu (approved):

PR #421, opened 2024-11-06
Add easy borrow form

yivlad:


packages/app/src/features/easy-borrow/components/form/Connector.tsx
```diff
@@ -0,0 +1,57 @@
+export interface ConnectorProps {
+  className?: string
+}
+
+export function Connector({ className }: ConnectorProps) {
+  return (
+    <svg className={className} preserveAspectRatio="xMidYMid meet" viewBox="0 0 84 93" fill="none">
```
EmilMalanczak, line 7: I think that should be exported as separate icon from assets

packages/app/src/ui/atoms/new/panel/Panel.tsx
```diff
@@ -34,4 +34,8 @@ const panelVariants = cva('rounded-sm', {
       m: 'p-3 md:p-8 sm:p-5',
     },
   },
+  defaultVariants: {
+    variant: 'primary',
+    spacing: 'm',
```
EmilMalanczak, line 39: Can you take a look at this [PR](https://github.com/marsfoundation/spark-app/pull/423/files#diff-b55c69118629554683b9630afb829f1746a414bc2135193f6a02cb136bc66899) Panel change? In current implementation we are missing `white` background panels which by design seems to be the primary ones. That PR will be kept as draft for a while so can we add something like this in here? That will reduce amount of possible regressions later on

EmilMalanczak:
it wasn't communicated anywhere but can you add mobile view as well? it available on mobile navbar [figma](https://www.figma.com/design/UPf7kTGjHBIvwjUKPGYcPd/Spark%3A-App?node-id=843-22966&t=JQQsXrlBJMcjN75H-4)

packages/app/src/ui/atoms/new/panel/Panel.tsx
oskarvu: `bg-primary` is white, `bg-primary-inverse` is black. I don't think we need to change color variants for panel component.

packages/app/src/features/easy-borrow/components/form/Connector.tsx
oskarvu: Yeah, you can put this in assets and then import it as a component like this: `import MoreIcon from '@/ui/assets/connector.svg?react'`

oskarvu (approved):

EmilMalanczak (approved):

PR #453, opened 2024-11-22
Unblock USA and remove VPN block

yivlad:


packages/app/.env.development
```diff
@@ -1,5 +1,7 @@
 # feature flags
-VITE_FEATURE_AUTH_IP_AND_ADDRESS_CHECKS=0
+VITE_FEATURE_AUTH_VPN=0
```
krzkaczor, line 2: Can be name then more appropriatly for example: `VITE_FEATURE_BLOCK_VPN`, `VITE_FEATURE_VERIFY_ADDRESS`, `VITE_FEATURE_BLOCK_REGIONS`

krzkaczor (approved):

PR #457, opened 2024-11-29
`common-universal` package

krzkaczor:
Note: running e2e tests require `common-universal` package built. I tried to fix it by setting `--conditions @marsfoundation/source` node options when running `playwright test` but got some weird error so for now i just made sure that the package is built. Let me know if this is an issue in practise.


packages/common-universal/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@marsfoundation/common-universal",
+  "version": "0.1.6",
+  "engines": {
+    "node": ">=18.0.0",
+    "pnpm": ">=8.0.0"
```
oskarvu, line 6: Can be removed. Root `package.json` has already engine specified for `9.14.2`

oskarvu (approved):

yivlad (approved):

PR #477, opened 2024-12-06
Remove colors not from design system

yivlad:


packages/app/src/ui/atoms/table/Table.tsx
```diff
@@ -40,7 +44,7 @@ const TableHead = React.forwardRef<HTMLTableCellElement, React.ThHTMLAttributes<
     <th
       ref={ref}
       className={cn(
-        'bg-white py-2 text-left align-middle font-medium text-muted-foreground [&:has([role=checkbox])]:pr-0',
+        'bg-white py-2 text-left align-middle font-medium text-secondary [&:has([role=checkbox])]:pr-0',
```
oskarvu, line 47: `font-medium` should be removed/replaced (probably in this PR, so there is no conflicts in second PR).

oskarvu (approved):

PR #485, opened 2024-12-10
Add mixpanel integration

yivlad:


packages/app/src/features/analytics/AnalyticsContainer.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const { showCookieBanner, onAcceptCookieBanner, onDeclineCookieBanner } = useAnalytics()
+
+  return (
+    <CookieBanner
+      onAccept={onAcceptCookieBanner}
+      onDecline={onDeclineCookieBanner}
+      className={cn(
+        'fixed inset-x-2 bottom-2 z-[1000] sm:right-6 sm:bottom-6 sm:left-auto',
+        'transform transition-all duration-500 ease-in-out',
+        showCookieBanner ? 'translate-y-0 opacity-100' : 'translate-y-full opacity-0',
```
krzkaczor, line 15: isn't it cleaner to simply `return null` if disabled?

```typescript
if (!showCookieBanner) {
	return null
}

return <CookieBanner ... />
```

packages/app/src/features/analytics/logic/useAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+  showCookieBanner: boolean
+  onAcceptCookieBanner: () => void
+  onDeclineCookieBanner: () => void
+}
+
+export function useAnalytics(): UseAnalyticsResult {
+  useTrackAccount()
+
+  const { accepted, setAccepted } = useStore((state) => state.analytics)
+  if (accepted) {
```
krzkaczor, line 16: honest question: won't this re-render and don't we need useEffect here?

packages/app/src/features/analytics/logic/useAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+    optInTracking()
+  } else {
+    optOutTracking()
+  }
+
+  const initialDataLoaded = useInitialDataLoaded()
```
krzkaczor, line 22: i don't get it, why is it needed?

packages/app/src/features/analytics/AnalyticsContainer.tsx
oskarvu: This way it won't animate nicely (sliding from down).

packages/app/src/features/analytics/logic/useAnalytics.ts
yivlad: It's deduplicated inside `optInTracking` function anyway

packages/app/src/features/analytics/logic/useAnalytics.ts
yivlad: The idea here is to drag user's attention to the cookie banner - it should appear after some time after the data is loaded, otherwise the user's attention might be switched to loaded content if it appears after the cookie banner

packages/app/src/features/analytics/AnalyticsContainer.tsx
yivlad: It'd done this way to have slide-in and slide-out animations

packages/app/src/domain/hooks/useWrite.ts
```diff
@@ -142,19 +142,13 @@ export function useWrite<TAbi extends Abi, TFunctionName extends ContractFunctio
...
-          } else {
-            recordEvent('tx-sent', {
-              walletType,
-              chainId,
-              receiver: parameters.request.address,
-              method: parameters.request.functionName,
-            })
-          }
+          trackEvent('tx-sent', {
+            walletType,
```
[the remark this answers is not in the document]
yivlad: Yep, this can be done but requires a bit of work to craft the board

packages/app/src/features/analytics/logic/useAnalytics.ts
krzkaczor: okay this makes sense but name is really confusing. IMO what you described here, should be put in a comment on that line ie. to catch user attention we display the cookie banner only after initial data is loaded.

PR #489, opened 2024-12-12
Add common-testnets package

yivlad:


packages/common-testnets/src/anvil/AnvilClient.ts
```diff
@@ -0,0 +1,61 @@
...
+      },
+      async revert(snapshotId: string) {
+        const result = await c.request({
+          method: 'evm_revert',
+          params: [snapshotId],
+        } as any)
+
+        assert(result === true, 'revert failed')
+
+        // anvil snapshot are "burned" after revert so we need to create a new one
```
krzkaczor, line 44:
```suggestion
        // anvil snapshots are "burned" after revert so we need to create a new one
```

packages/common-testnets/src/TestnetFactory.ts
```diff
@@ -0,0 +1,19 @@
+import { TestnetClient } from './TestnetClient'
+
+export interface TestnetCreateResult {
+  client: TestnetClient
+  cleanup: () => Promise<void>
+}
+
+export interface TestnetFactory {
```
krzkaczor, line 8: Please add a block comment here explaining that TestnetFactories will normalize the tip of the blockchain which results in extra blocks and slight timestamp difference.

.github/workflows/ci.yml
```diff
@@ -65,7 +65,8 @@ jobs:
       - run: pnpm install
 
       - run: pnpm check
-      - run: pnpm verify # does linting, type checking, and tests in parallel
+      - name: Verify # verifies all packages except common-testnets
+        run: pnpm run --parallel --aggregate-output --reporter append-only --filter './packages/{app,common-nodejs,common-universal}' verify
```
krzkaczor, line 69: i think we can simplify this. So the mental model should be that `verify` is fast (linting, unit tests etc) and we can always run in on every single project. In case of `common-testnets` verify would run eslint and typecheck but no tests because these are not unit tests but rather integration/e2e tests. So the task should be renamed to test-e2e and then the CI job for testnets runs only that.

krzkaczor (approved):

PR #491, opened 2024-12-13
Refactor e2e tests setup

yivlad:
Closes FRO-1189


packages/app/src/test/e2e/testnet-cache.ts
```diff
@@ -0,0 +1,59 @@
...
+    const keys = Array.from(testnetCache.keys())
+    for (const key of keys) {
+      const context = testnetCache.get(key)!
+      testnetCache.delete(key)
+      return context.cleanup()
+    }
+  })
+}
+
+export async function getTestnetContext({
```
krzkaczor, line 23: this file is IMO not about caching but mostly about testnet creation so i would call it getTestnetContext and move this function above caching stuff.

packages/app/src/test/e2e/setup.ts
```diff
@@ -43,122 +47,142 @@ export type AccountOptions<T extends ConnectionType> = T extends 'not-connected'
...
 
-  await page.goto(buildUrl(options.initialPage, options.initialPageParams))
+  async function incrementTime(seconds: number): Promise<void> {
+    await updateBrowserAndNextBlockTime(seconds)
+    await testnetClient.mineBlocks(1n)
+    await updateBrowserAndNextBlockTime(5)
+  }
+
+  // @note: Sync time in browser with current time on blockchain,
+  // set next block to be mined timestamp to be 5 seconds more.
```
krzkaczor, line 95: I think the second line of this comment in redundant.

packages/app/src/test/e2e/setup.ts
```diff
@@ -43,122 +47,142 @@ export type AccountOptions<T extends ConnectionType> = T extends 'not-connected'
...
 
-export type SetupReturn<T extends ConnectionType> = T extends 'not-connected'
-  ? {
-      getLogs: () => string[]
-    }
-  : {
-      account: Address
-      getLogs: () => string[]
-    }
+export type SetupReturn<T extends ConnectionType> = (T extends 'not-connected' ? {} : { account: Address }) & {
```
krzkaczor, line 62: This should be called imo something like `TestContext`. Then functions manipulating the state of the simulation might be under `TestController` scope:
`testContext.controller.incrementTime` etc.

packages/app/src/test/e2e/testnet-cache.ts
krzkaczor: what about renaming this file?

oskarvu: @krzkaczor Renamed.

packages/app/src/test/e2e/setup.ts
```diff
@@ -43,122 +41,152 @@ export type AccountOptions<T extends ConnectionType> = T extends 'not-connected'
...
-      getLogs: () => string[]
-    }
+export type ProgressSimulation = (seconds: number) => Promise<void>
+export interface TestnetController {
+  client: TestnetClient
+  progressSimulation: ProgressSimulation
+  progressSimulationAndMine: ProgressSimulation
+}
+
+export type TestContext<T extends ConnectionType> = (T extends 'not-connected' ? {} : { account: Address }) & {
```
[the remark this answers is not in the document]
yivlad: It's kinda convenient that if you need an address you don't have to make an assertion. What about making `T` equal `not-connected` by default? As in here https://github.com/marsfoundation/spark-app/pull/491/commits/53a3a24051abc329f3c3cb98d1b541889b6e5ba9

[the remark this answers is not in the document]
krzkaczor: > It's kinda convenient that if you need an address you don't have to make an assertion.

But you need to make sure to pass appropriate generic type, no? Doesn't it result in some hard cast? or some other "unsafenest"? But I defer to your opinion here because you actually used this type xD

PR #494, opened 2024-12-16
Add constant padding to random hexes, hashes and addresses

barrutko:


packages/common-universal/src/types/CheckedAddress.ts
```diff
@@ -19,10 +19,11 @@ export function CheckedAddress(value: string): CheckedAddress {
  * Generates a random address. Tries to represent desired ascii prefix as hex value. Helps to identify transactions in the logs.
  */
 CheckedAddress.random = (asciiPrefix = ''): CheckedAddress => {
+  const constantAddressPrefix = '0000'
```
krzkaczor, line 22: i thought about this more: i think a good prefix is `deadbeef` as it's obviously not random

barrutko: changed to deadbeef

packages/common-universal/src/types/CheckedAddress.ts
```diff
@@ -19,10 +19,11 @@ export function CheckedAddress(value: string): CheckedAddress {
  * Generates a random address. Tries to represent desired ascii prefix as hex value. Helps to identify transactions in the logs.
  */
 CheckedAddress.random = (asciiPrefix = ''): CheckedAddress => {
```
oskarvu, line 21: I guess it will be `infix` now, since it's not first part of the address anymore.

I'm curious, why this change was needed in the first place?

barrutko: I don't know if we need to be that precise to rename that to `infix` - both of these parts are still `pre`randomPart :P

yivlad (approved):

oskarvu (approved):

PR #499, opened 2024-12-18
Rewrite easy borrow e2e tests

oskarvu:


packages/app/src/config/feature-flags/appConfig.default.ts
```diff
@@ -49,6 +49,14 @@ export function getDefaultAppConfig(): AppConfig {
             address: '0xdC035D45d973E3EC169d2276DDab16f1e407384F',
             decimals: 18,
           },
+          sUSDS: {
```
yivlad, line 52: I think this makes sandbox creation seed user's wallet with sUSDS, and we don't want to seed savings tokens in sandbox mode

packages/app/src/pages/Borrow.PageObject.ts
```diff
@@ -1,176 +1,168 @@
...
+
+  async expectSuccessPage({
+    deposited,
+    borrowed,
+    testnetClient,
+    assetsWorthOverride,
+  }: {
+    deposited: TestTokenWithValue[]
+    borrowed: TestTokenWithValue
+    testnetClient: TestnetClient
```
yivlad, line 140: Can be fetched from `this.testContext`. Also can we explore removing the data fetching and just hardcoding the USD values?

PR #525, opened 2025-01-07
Fix package versions to 0.0.1 

krzkaczor:
* since packages are automatically versioned, i set version to something that is more neutral. LMK if you have better ideas.
* also uncomments paths since tests are stable now


.github/workflows/ci-common-testnets.yml
```diff
@@ -2,10 +2,9 @@ name: CI (common-testnets)
 
 on:
   pull_request:
-    # @todo: uncomment when testnet factories are stable
-    # paths:
-    #   - "packages/common-testnets/**"
-    #   - "pnpm-lock.yaml"
+    paths:
+      - "packages/common-testnets/**"
+      - "pnpm-lock.yaml"
```
yivlad, line 7:
```suggestion
      - ".github/workflows/ci-common-testnets.yml"
```
This is more consistent with other jobs and I think actually makes more sense

yivlad (approved):

oskarvu (approved):

PR #526, opened 2025-01-07
Make markets TVL extract proportioned available DAI from D3M

yivlad:


packages/app/src/features/markets/logic/aggregate-stats.ts
```diff
@@ -24,10 +24,16 @@ export function aggregateStats(marketInfo: MarketInfo, D3MInfo: D3MInfo | undefi
   )
   const totalAvailableUSD = aggregatedValues.totalLiquidityUSD.minus(aggregatedValues.totalDebtUSD)
   const daiReserve = marketInfo.findReserveByToken(marketInfo.DAI)
-  const totalValueLockedUSD =
-    D3MInfo && daiReserve
-      ? NormalizedUnitNumber(totalAvailableUSD.minus(D3MInfo.D3MCurrentDebtUSD.minus(daiReserve.totalVariableDebtUSD)))
-      : undefined
+  const daiAvailable = daiReserve
+    ? NormalizedUnitNumber(daiReserve.totalLiquidity.minus(daiReserve.totalVariableDebtUSD))
```
oskarvu, line 28: Since we are working with usd values here, I think we should go with `daiReserve.totalLiquidityUSD`. It won't assume fixed oracle price this way.

oskarvu (approved):

PR #527, opened 2025-01-08
Bump biome

krzkaczor:
Bump biome and reformat all files. There is more changes that I thought but I guess its "for the best".

I need this to finish FRO-1215.

### DIFF
--- a/packages/app/src/features/markets/components/markets-table/components/AssetNameCell.tsx
+++ b/packages/app/src/features/markets/components/markets-table/components/AssetNameCell.tsx
@@ -26,7 +26,7 @@ export function AssetNameCell({ token, reserveStatus, 'data-testid': dataTestId 
       {tokenImage && (
         <div className="flex shrink-0">
           <ColorFilter variant={isPaused ? 'red' : 'none'}>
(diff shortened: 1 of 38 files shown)


packages/app/src/features/markets/components/markets-table/components/AssetNameCell.tsx
```diff
@@ -26,7 +26,7 @@ export function AssetNameCell({ token, reserveStatus, 'data-testid': dataTestId
       {tokenImage && (
         <div className="flex shrink-0">
           <ColorFilter variant={isPaused ? 'red' : 'none'}>
-            <TokenIcon token={token} className="flex h-6 w-6 lg:h-6 md:h-5 lg:w-6 md:w-5" />
+            <TokenIcon token={token} className="flex h-6 w-6 md:h-5 md:w-5 lg:h-6 lg:w-6" />
```
nezouse, line 29: Nothing to change here, but instead of using` h-5 w-5` in the future it can be `size-5`

nezouse (commented):
Sometimes active/focus states works weird with specificity so i would double check in app that it works after sorting, but otherwise lgtm

yivlad (approved):

krzkaczor:
@nezouse i reviewed the app manually. Looks good. We also have chromatic visual regressions tests.

PR #528, opened 2025-01-08
Use numbers types from common-universal

yivlad:


packages/common-universal/src/types/UnixTime.ts
```diff
@@ -7,7 +7,11 @@ import { Opaque } from './Opaque.js'
  * Can be used to represent Unix timestamps. i.e. 1729691675
  */
 export type UnixTime = Opaque<bigint, 'UnixTime'>
-export function UnixTime(value: NumberLike): UnixTime {
+export function UnixTime(value: NumberLike | Date): UnixTime {
+  if (value instanceof Date) {
```
krzkaczor, line 11: Hmm we already have `UnixTime.fromDate`. Either it should use `UnixTime.fromDate` under the hood, or it should be deleted.

krzkaczor: Also, remember now if we add new functionality to common primitives we should add accompanying tests (but it's tested already).

packages/app/src/domain/market-info/math.test.ts
```diff
@@ -1,4 +1,4 @@
-import { bigNumberify } from '@/utils/bigNumber'
+import { bigNumberify } from '@marsfoundation/common-universal'
```
oskarvu, line 1: We should import stuff from same package in one line. Happens in multiple places in this PR.

krzkaczor: tbh i don't mind it that much. Soon biome will autofix such errors...

oskarvu (approved):

PR #535, opened 2025-01-11
Fix promise logging

krzkaczor:
Before, promises were logged as `{}`. Now it should be easier to spot such mistakes and fix the logging. 

The best would be to simply prevent from passing `Promise` instances to logger but I don't think it's doable right now...

### DIFF
--- a/packages/common-nodejs/src/logger/Logger.test.ts
+++ b/packages/common-nodejs/src/logger/Logger.test.ts
@@ -80,6 +80,25 @@ describe(Logger.name, () => {
     expect(transport.log).toHaveBeenOnlyCalledWith(lines.join(''))
   })
 
+  it('marks promised values in pretty output', () => {
+    const transport = createTestTransport()
+    const logger = new Logger({
+      transports: [
+        {
+          transport: transport,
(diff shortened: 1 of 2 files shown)


nezouse (approved):

packages/common-nodejs/src/logger/Logger.test.ts
```diff
@@ -80,6 +80,25 @@ describe(Logger.name, () => {
     expect(transport.log).toHaveBeenOnlyCalledWith(lines.join(''))
   })
 
+  it('marks promised values in pretty output', () => {
+    const transport = createTestTransport()
+    const logger = new Logger({
+      transports: [
+        {
+          transport: transport,
```
oskarvu, line 88:
```suggestion
          transport,
```

oskarvu (approved):

packages/common-nodejs/src/logger/Logger.test.ts
krzkaczor: This would be haved to be fixed in the whole file, together probably with some lint rule... I will ignore this for now.

PR #540, opened 2025-01-14
Fix toString on BigNumber using scientific notation

nezouse:


packages/common-universal/src/math/bigNumber.ts
```diff
@@ -2,6 +2,8 @@ import BigNumber from 'bignumber.js'
 
 import { assert, BaseUnitNumber } from '../index.js'
 
+BigNumber.config({ EXPONENTIAL_AT: 1e9 })
```
oskarvu, line 5: Can you go with something like this, so It never returns exponential notation:
`BigNumber.config({ EXPONENTIAL_AT: [-Infinity, Infinity] })`

Going with 1 billy might not be enough.

oskarvu: Hmm, looking at tests now and it seems that 1e9 is not 1 billion, but more.

oskarvu (approved):

packages/common-universal/src/math/bigNumber.ts
krzkaczor: Yup, i think this needs a comment. This means (from my understanding) that it would try to use scientific notation only when exponent `e` component of a number is `1e9`. `e` value goes to `1000000000` (pic). So, as docs say, in practise it should never happen. I guess my question is can we set it to `infinity` 🤔 


image

nezouse, line 5:
```suggestion
// Only use scientific notation if number's exponent is greater than 1e9 (so 1e9 decimals of precision)
BigNumber.config({ EXPONENTIAL_AT: 1e9 })
```

PR #547, opened 2025-01-16
Actions master hook

oskarvu:
Closes FRO-1226

### DIFF
--- a/packages/app/src/features/actions/logic/useCanWalletBatch.ts
+++ b/packages/app/src/features/actions/logic/useCanWalletBatch.ts
@@ -1,0 +1,64 @@
+import { SupportedChainId } from '@/config/chain/types'
+import { useOriginChainId } from '@/domain/hooks/useOriginChainId'
+import { QueryKey, UseQueryResult, queryOptions, skipToken, useQuery } from '@tanstack/react-query'
+import { Address } from 'viem'
+import { Config, useAccount, useConfig } from 'wagmi'
+import { getCapabilities } from 'wagmi/actions/experimental'
+
+const CAPABILITIES_QUERY_FN_TIMEOUT = 3000
+
+export function useCanWalletBatch(): UseQueryResult<boolean, Error> {
+  const config = useConfig()
+  const chainId = useOriginChainId()
+  const { address: account } = useAccount()
+  return useQuery(canWalletBatchQueryOptions({ config, chainId, account }))
+}
+
+export interface CanWalletBatchQueryOptionsParams {
+  account: Address | undefined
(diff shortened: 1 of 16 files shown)


packages/app/src/features/actions/logic/useCanWalletBatch.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export interface CanWalletBatchQueryOptionsParams {
+  account: Address
+  chainId: SupportedChainId
+  config: Config
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export function canWalletBatchQueryOptions({ account, chainId, config }: CanWalletBatchQueryOptionsParams) {
+  return queryOptions({
```
yivlad, line 27: Does it need a custom retry function or number? If this throws, wouldn't that mean that by default react-query will retry 3 times possibly taking quite a long time?

packages/app/src/features/topbar/logic/usePrefetchCanWalletBatchQuery.ts
```diff
@@ -0,0 +1,16 @@
+import { useOriginChainId } from '@/domain/hooks/useOriginChainId'
+import { canWalletBatchQueryOptions } from '@/features/actions/logic/useCanWalletBatch'
+import { useQueryClient } from '@tanstack/react-query'
+import { useAccountEffect, useConfig } from 'wagmi'
+
+export function usePrefetchCanWalletBatchQuery(): void {
+  const queryClient = useQueryClient()
+  const chainId = useOriginChainId()
```
yivlad, line 8: Why origin chainId?

oskarvu: Was doing some tests while on sandbox and forgot to switch back. Thanks!

packages/app/src/features/actions/logic/useCanWalletBatch.ts
oskarvu: I've handled fetching/prefetching a bit differently.

oskarvu: Since this is suspended query, error is handled inside query function, so it never will be retried. In addition, there is prefetch done in other place, so there are 2 calls in total.

I noticed two situations in which errors can happen:
- when wallet isn't supporting batching -> false is returned from query fn
- sometimes wallet connect has lags and doesn't respond at all -> timeout is triggered and false returned

yivlad (approved):

PR #550, opened 2025-01-19
Fix common packages package.json

krzkaczor:
* testnet package had wrong `peerDependencies`
* drop canary naming, simply use latest


.github/workflows/release-canary-common-nodejs.yml
oskarvu, line 1: I guess name of the file can now be: `release-latest-common-nodejs`.

oskarvu (approved):

yivlad (approved):

PR #555, opened 2025-01-22
Add deposit call to action component

oskarvu:
Closes FRO-1235


packages/app/src/features/savings/components/deposit-cta-panel/DepositCTAPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+      action: () => void
+    }
+    secondary: {
+      title: 'Try in Sandbox'
+      action: () => void
+    }
+  }
+}
+
+export function DepositCTAPanel({ inputTokens, outputToken, globalStats, description, actions }: DepositCTAPanelProps) {
```
yivlad, line 36: I'd rename this component to something that stresses that this is visible only when hasn't connected yet

yivlad (approved):

packages/app/src/features/savings/components/deposit-cta-panel/DepositCTAPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <Panel
+      spacing="m"
+      className={cn(
+        'grid grid-cols-1 gap-8 md:grid-cols-[3fr_2fr]',
+        panelBgVariants({ bg: outputToken.symbol.toLowerCase() }),
```
yivlad, line 42: Does it work when symbol is none of the given?

packages/app/src/features/savings/components/deposit-cta-panel/DepositCTAPanel.tsx
oskarvu: It's also visible, when user is connected, but doesn't have any funds deposited yet.

packages/app/src/features/savings/components/deposit-cta-panel/DepositCTAPanel.tsx
oskarvu: Good question, I assumed it will, but it won't. Let me fix that.

PR #556, opened 2025-01-22
Add connected main panel group to savings

yivlad:


packages/app/src/features/savings/components/account-main-panel-group/utils.ts
oskarvu, line 1: There is very similar util at deposit-cta-panel. We should extract and merge them into one. Same with AccountType.

packages/app/src/features/savings/components/account-main-panel-group/AccountMainPanelGroup.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const mainPanelVariants = cva('bg-cover bg-right bg-no-repeat', {
+  variants: {
+    bg: {
+      susds: 'bg-[url(/src/ui/assets/savings/accounts/usds-bg.svg)] bg-primary-inverse',
+      susdc: 'bg-[url(/src/ui/assets/savings/accounts/usdc-bg.svg)] bg-primary-inverse',
+      sdai: 'bg-[url(/src/ui/assets/savings/accounts/dai-bg.svg)] bg-primary-inverse',
+    },
+  },
+  defaultVariants: {
```
oskarvu, line 117: nit: we don't need default variant here since function to get variant will return default anyway.

oskarvu (approved):

PR #557, opened 2025-01-22
Report query and mutation errors to sentry

oskarvu:
Closes FRO-1254

- We watched query cache for mutation and query errors
- But those errors were filtered out in beforeSend at sentry init

We didn't receive those errors for months...

Also removes problematic sentry plugin.


packages/app/src/features/actions/logic/useBatchWrite.ts
```diff
@@ -95,6 +97,12 @@ export function useBatchWrite({ contracts, enabled = true, callbacks = {} }: Use
     return { kind: 'ready' }
   })()
 
+  useOnDepsChange(() => {
+    if (status.kind === 'error') {
+      captureError(new HandledWriteError(status.error))
```
yivlad, line 102: Can't we wrap the error in query cache observer? I'd rater prefer it to the hook

yivlad: Another quick idea: maybe we should leverage the hints object in beforeSend function in sentry? For instance we could say that some errors (for instances ones form react query) we want to push no matter what

oskarvu: Wrapping error in query cache observer was the first step I took, but I noticed a lot of unwanted errors being reported. I will run a test to show you how many errors we don't care about are being listed. Good example can be a wallet capabilities check - legitimate error that gives us answer if wallet supports batching. In such setup also this error would be reported and we would need add more filters.

At the same time I understand concerns that some of valuable errors might not be reported when we capture them only on the hook level.

Let's try to capture errors on query cache level, adding some additional filters for some unwanted errors and see how it goes.

yivlad (approved):

PR #564, opened 2025-01-23
Add deposit to sUSDC actions

yivlad:


packages/app/src/features/actions/flavours/deposit-to-savings/logic/depositToSavingsAction.ts
```diff
@@ -52,6 +52,8 @@ export function createDepositToSavingsActionConfig(
             args: [assetsAmount, account, SPARK_UI_REFERRAL_CODE],
           })
         case 'dai-to-sdai':
+        case 'usdc-to-susdc':
```
oskarvu, line 55: I wonder if we shouldn't use usdcVault abi and deposit function where we also pass `minShares`. If we use simpler version, `minShares` is set for `0`.

packages/app/src/features/actions/flavours/deposit-to-savings/logic/depositToSavingsActionBase.test.ts
```diff
@@ -159,15 +159,20 @@ describe(createDepositToSavingsActionConfig.name, () => {
...
         handlers.contractCall({
           to: getContractAddress(usdcVaultAddress, chainId),
           abi: usdcVaultAbi,
           functionName: 'deposit',
-          args: [toBigInt(usdc.toBaseUnit(depositValue)), account],
+          args: [
+            toBigInt(usdc.toBaseUnit(depositValue)),
+            account,
+            toBigInt(susdc.toBaseUnit(minAmountOut)),
+            SPARK_UI_REFERRAL_CODE,
```
oskarvu, line 174: Can we double check on this? In SC fallback value is `0`, and this contract are developed by us, so maybe we are `0` in this context?

oskarvu (approved):

PR #565, opened 2025-01-24
Add usdc account withdraw and send actions

oskarvu:
Closes FRO-1243


packages/app/src/features/actions/flavours/withdraw-from-savings/logic/withdrawFromSavingsAction.ts
```diff
@@ -156,6 +157,38 @@ export function createWithdrawFromSavingsActionConfig(
...
+            return ensureConfigTypes({
+              address: savingsToken.address,
+              abi: usdcVaultAbi,
+              functionName: 'redeem',
+              args: [argsAmount, receiver, account, toBigInt(minAssetsOut)],
+            })
+          }
+
+          // @note: Assumes no psm fees
+          const maxSharesIn = action.savingsToken.toBaseUnit(
```
yivlad, line 178: Will have the same problem as `base-usdc-to-susds` action - should use `formatMaxAmountInForPsm3`. Probably should add a unit test for this

yivlad (approved):

packages/app/src/features/actions/flavours/withdraw-from-savings/logic/withdrawFromSavingsAction.ts
oskarvu: Let's handle such things on savings info level, as @yivlad suggested.

PR #568, opened 2025-01-24
Add support for more arguments in min function

nezouse:
closes TS-93

### DIFF
--- a/packages/common-universal/src/types/NormalizedUnitNumber.ts
+++ b/packages/common-universal/src/types/NormalizedUnitNumber.ts
@@ -17,6 +17,17 @@ NormalizedUnitNumber.toBaseUnit = function toBaseUnit(value: NormalizedUnitNumbe
   return BaseUnitNumber(lowerPrecisionNumber.shiftedBy(decimals))
 }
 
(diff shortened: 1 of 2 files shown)


packages/common-universal/src/types/NormalizedUnitNumber.ts
```diff
@@ -17,17 +18,13 @@ NormalizedUnitNumber.toBaseUnit = function toBaseUnit(value: NormalizedUnitNumbe
...
-): NormalizedUnitNumber {
-  let min = a
-  const values = [b, ...rest]
-  for (const value of values) {
-    if (value.lt(min)) {
-      min = value
+NormalizedUnitNumber.min = function min(...values: NormalizedUnitNumber[]): NormalizedUnitNumber {
+  assert(values.length > 0, 'Requires at least 1 arg')
+  let min = values[0]
+  for (let i = 1; i < values.length; i++) {
```
krzkaczor, line 24: nit: alternatively you can simply for each every value and simply the first value will be always ignored (because it's not less than `min`)....

krzkaczor (approved):

PR #569, opened 2025-01-24
Add USDC Savings config to mainnet and base

yivlad:
As USDC vault on Base uses PSM3 under the hood, appropriate actions should take care of rounding for USDC. Couldn't be done on the level of SavingInfo because most often worth of USDC deposited is bigger than the worth of sUSDC received, but when withdrawing worth of USDC withdrawn is smaller than worth of sUSDC burnt. Therefore unambiguous worth of USDC cannot be established without the context of operation.


packages/app/src/features/actions/flavours/withdraw-from-savings/logic/withdrawFromSavingsAction.ts
```diff
@@ -189,6 +188,39 @@ export function createWithdrawFromSavingsActionConfig(
...
+              address: savingsToken.address,
+              abi: usdcVaultAbi,
+              functionName: 'redeem',
+              args: [argsAmount, receiver, account, toBigInt(minAssetsOut)],
+            })
+          }
+
+          // @note: Assumes no psm fees
+          const maxShares = formatMaxAmountInForPsm3({
+            susds: action.savingsToken,
```
oskarvu, line 209: Savings token will be susdc here, but technically, it's just a wrapper on susds, so maybe we don't need to change this.

oskarvu (approved):

PR #570, opened 2025-01-24
Bring back E2E tests sharding

yivlad:


.github/workflows/ci.yml
```diff
@@ -127,7 +128,7 @@ jobs:
 
       - run: pnpm exec playwright install --with-deps chromium
         working-directory: ./packages/app
-      - run: pnpm run test-e2e
+      - run: pnpm run test-e2e --shard ${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
```
krzkaczor, line 131: IMO this, and (matrix definition) should be only changes done to enable sharding. No reports merging, if something fails you would download a report from a particular shard. 

This should basically, linearly scale `run test-e2e` step that takes almost 6m. Ofc each runner would still spend around 2m on the setup but this should still speed things up. And ofc some e2e tests might take longer to execute so some shards might be slower but def some speed up should be observed.

image

oskarvu (approved):

krzkaczor (approved):

PR #572, opened 2025-01-27
Use one chart component for multiple savings

oskarvu:
Closes FRO-1247


packages/app/src/features/savings/views/GuestView.tsx
```diff
@@ -50,7 +50,7 @@ export function GuestView({
           savingsMeta={savingsMeta}
           openSandboxModal={openSandboxModal}
         />
-        {displaySavingsChart && <Charts {...savingsChartsInfo} />}
+        {displaySavingsChart && <SavingsCharts {...chartsMeta} {...savingsChartsInfo} />}
```
yivlad, line 53: I'm wondering maybe we should integrate `chartsMeta` into `savingsChartsInfo`? This way the tooltip text will be decided on in master hook

oskarvu: Good idea 👍🏻

oskarvu: I looked at it again, and `useSavingsChartsInfoQuery` is not parametrized by savings token. It just fetches all savings charts data at once. We need create ticket to refactor this hook.

In the meantime I added util to get meta using symbol of savings token (temporary solution, backwards compatible).

yivlad (approved):

PR #576, opened 2025-01-29
Add parametrized savings account view

oskarvu:
Closes FRO-1248


packages/app/src/features/savings/views/SavingsAccountView.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { MigrationInfo } from '../logic/makeMigrationInfo'
+import { SavingsAccountEntryAssets, SavingsTokenDetails } from '../logic/useSavings'
+
+export interface SavingsAccountViewProps {
+  savingsTokenDetails: SavingsTokenDetails
+  entryAssets: SavingsAccountEntryAssets[]
+  mostValuableAsset: TokenWithBalance
+  assetsConvertSupported: boolean
+  savingsChartsInfo: UseSavingsChartsInfoQueryResult
+  openDialog: OpenDialogFunction
```
yivlad, line 22: Pass functions like `openSavingsDepositDialog: (token: Token) => {}`?

oskarvu: I'm not sure. I still need to pass `openDialog` to `EntryAssetsPanel` (dialog is opened there depending on which button user clicks, so I don't know input token ahead of time). I can add more specific functions, like: `openSavingsDepositMostValuableToken`, `openSavingsWithdraw` and `openSavingsSend`, but I think I prefer to just reuse `openDialog` here. What do you think?

oskarvu: Let's abstract this when we delete old views and refactor dialogs.

yivlad (approved):

PR #590, opened 2025-02-04
Add new savings accounts banner

yivlad:


oskarvu:
Can we disable button effect when on savings page? So on savings it's just a static banner (no cursor pointer and no hover effect).

oskarvu (approved):

PR #591, opened 2025-02-05
Add savings usdc account related tests

oskarvu:
Closes FRO-1266


packages/app/src/features/dialogs/savings/deposit/logic/createTxOverview.ts
```diff
@@ -37,9 +38,7 @@ export function createTxOverview({
   })
 
   return {
-    baseStable:
-      (savingsToken.symbol === tokensInfo.sDAI?.symbol ? tokensInfo.DAI : tokensInfo.USDS) ??
-      raise('Cannot find stable token'),
+    underlyingToken: findUnderlyingToken(savingsToken, tokensInfo) ?? raise('Cannot find underlying token'),
```
yivlad, line 41: Pass SavingsAccount from SavingsAccountRepository instead, it will have the savingsConverter, savingsToken and underlyingToken

yivlad: I don't think we should use `findUnderlyingToken` function anywhere as it's just really part of the account config

packages/app/src/features/dialogs/savings/withdraw/e2e/base/usdc-account/SendUSDC.test-e2e.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const receiver = randomAddress('bob')
+  const amount = 7000
+  const usdc = TOKENS_ON_FORK[base.id].USDC
+
+  test.beforeEach(async ({ page }) => {
+    const testContext = await setup(page, {
+      blockchain: {
+        chain: base,
+        blockNumber: BASE_MOCK_SUSDC_ACTIVE_BLOCK_NUMBER,
```
yivlad, line 21: I think it can be just `BASE_SUSDC_ACTIVE_BLOCK_NUMBER`

packages/app/src/features/dialogs/savings/withdraw/e2e/base/usdc-account/WithdrawMaxUSDC.test-e2e.ts
```diff
@@ -0,0 +1,76 @@
...
+  test('displays transaction overview', async () => {
+    await withdrawDialog.expectNativeRouteTransactionOverview({
+      routeItems: [
+        {
+          tokenAmount: '9,666.53 sUSDC',
+          tokenUsdValue: '$10,000.00',
+        },
+        {
+          tokenAmount: '9,999.99 USDC',
+          tokenUsdValue: '$9,999.99',
```
yivlad, line 58: Do you know why the difference is 1 cent? I think the contract is written the amount may be less the withdrawn amount by 1 decimal point, e.g. 6 zeros in case of USDC Wondering what may cause the 1 cent difference

packages/app/src/test/e2e/injectSetup.ts
```diff
@@ -71,3 +75,31 @@ export async function injectNetworkConfiguration({
...
+      if (isSusdcDeployed) {
+        ;(window as any)[PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE_KEY] = true
+      }
+    },
+    { PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE_KEY, isSusdcDeployed },
+  )
+}
+
+async function isSudcDeployed(testnetClient: TestnetClient, chainId: number): Promise<boolean> {
+  const susdsAddress = (() => {
```
yivlad, line 93:
```suggestion
  const susdcAddress = (() => {
```

packages/app/src/features/dialogs/savings/withdraw/e2e/base/usdc-account/SendUSDC.test-e2e.ts
oskarvu: I think we can leave it, as it notes that is not a final version of susdc. I had in mind, that after we have final version deployed, we will change this constants names and also update block number value.

packages/app/src/features/dialogs/savings/withdraw/e2e/base/usdc-account/WithdrawMaxUSDC.test-e2e.ts
oskarvu: This is because how we handle form max amount. It uses token.toUsd under the hood, and price returned gives such value. This problem seems to be only surfaced in sandbox mode. I'll crate a [ticket](https://linear.app/phoenixlabs/issue/FRO-1277/check-withdraw-input-max-value-handling) to investigate this.

PR #593, opened 2025-02-06
Fill general stats bar with api data

oskarvu:
Closes FRO-1246


packages/app/src/domain/common/validation.ts
```diff
@@ -16,6 +16,7 @@ export const numberLikeSchema = z.union([
 export const percentageSchema = numberLikeSchema.transform((a) => Percentage(a))
 export const percentageAboveOneSchema = numberLikeSchema.transform((a) => Percentage(a, { allowMoreThan1: true }))
 export const normalizedUnitNumberSchema = numberLikeSchema.transform((a) => NormalizedUnitNumber(a))
+export const normalizedNumberSchema = numberLikeSchema.transform((a) => NormalizedUnitNumber(a).toNumber())
```
yivlad, line 19: Why not just use `normalizedUnitNumberSchema` everywhere instead? This would be more consistent with what we do in other parts of the cdoe

yivlad (approved):

PR #594, opened 2025-02-06
Display apy data when user has funds deposited

oskarvu:
Closes FRO-1278


packages/app/src/features/savings/components/account-main-panel-group/AccountMainPanelGroup.tsx
```diff
@@ -22,12 +22,14 @@ export interface AccountMainPanelGroupProps {
   openSendDialog: () => void
   openWithdrawDialog: () => void
   projections: Projections
```
yivlad, line 24: I guess we can also remove `thirtyDays` from projections object

yivlad (approved):

PR #597, opened 2025-02-06
Fix solid fetch to finish after 5 retries

nezouse:


packages/common-universal/src/networking/solidFetch.test.ts
```diff
@@ -0,0 +1,10 @@
+import { expect } from 'earl'
+import { solidFetch } from './solidFetch.js'
+
+describe(solidFetch.name, () => {
```
nezouse, line 4: Great features of this test:
- it lasts 5 seconds
- it logs a bunch of bullshit
- it will timeout whole test suite if something fails the same way before my changes

krzkaczor (approved):

yivlad (approved):

PR #608, opened 2025-02-10
Add usdc account feature flag

oskarvu:
Closes FRO-1286


packages/app/src/config/chain/index.ts
```diff
@@ -47,6 +47,7 @@ const commonTokenSymbolToReplacedName = {
 
 const PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE =
   import.meta.env.VITE_PLAYWRIGHT !== '1' || (window as any)[PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE_KEY] === true
+const USDC_ACCOUNT_ENABLED = PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE && import.meta.env.VITE_FEATURE_USDC_ACCOUNT === '1'
```
yivlad, line 50: Now that I'm looking at this it's a bit misleading, maybe we should do something like

```suggestion
const PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE = (window as any)[PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE_KEY] === true
const USDC_ACCOUNT_ENABLED = PLAYWRIGHT_SUSDC_CONTRACTS_AVAILABLE || (import.meta.env.VITE_PLAYWRIGHT !== '1' && import.meta.env.VITE_FEATURE_USDC_ACCOUNT === '1')
```

yivlad (approved):

PR #619, opened 2025-02-13
Add external link to general stats bar

oskarvu:
Closes FRO-1297


packages/app/src/features/savings/components/general-stats-bar/GeneralStatsBar.tsx
```diff
@@ -43,6 +47,11 @@ export function GeneralStatsBar({ accountSavingsToken, generalStatsResult }: Gen
           )}
         </Value>
       </Stat>
+      <Stat>
+        <Link to={links.skyInfoSavingsDashboard} variant="decorator" className="flex" external>
```
yivlad, line 51: Why not LinkDecorator?

oskarvu: So it's wrapped in anchor tag, which is more semantically correct. And it doesn't give me styles for focus.

yivlad (approved):

PR #665, opened 2025-02-26
Handle rewards page empty states

oskarvu:
Closes FRO-1339


packages/app/src/ui/molecules/icon-stack/IconStack.tsx
```diff
@@ -68,10 +67,32 @@ const iconVariants = cva('rounded-full', {
...
-      true: 'box-content border-2',
+    border: {
+      white: 'border-base-white',
+      transparent: 'border-transparent',
     },
   },
+  compoundVariants: [
+    {
+      border: ['white', 'transparent'],
+      className: 'box-content',
```
yivlad, line 78: I'm wondering why we need this? Is border set somewhere apart from here? If yes what's the difference between this and border-0?

yivlad (approved):

PR #666, opened 2025-02-26
Preconfigure rewards when entering sandbox

yivlad:


oskarvu:
Just a note: mocked rewards don't survive refresh. I guess it's not a problem since we use them for development/demo purposes.

oskarvu (approved):

PR #672, opened 2025-03-03
Move HttpClient to common-nodejs

b-tarczynski:
The logger is in `common-nodejs`, so I added HttpClient there instead of universal. I talked with Vlad about usage of `solidFetch` on frontend and it's not needed as there are already better retires from tanstack query


nezouse:
@krzkaczor logger is node-only package so it's not in universal, wdyt?

krzkaczor:
@b-tarczynski  @nezouse  got it. 

Hmmmm okay, i think in the short term it's fine to make it node specific, but in the longer term I believe we should make logger universal (or at least, create a thin wrapper on console.log which would be interface compatible). 

This way we could even use `common-testnets` package to create sandbox mode forks in browser (once we make lambdas interface compatible...). CC: @yivlad

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 dist
+.idea
```
krzkaczor, line 3: add this to your global git ignore file please.

krzkaczor (approved):

PR #677, opened 2025-03-04
Add tests for HttpClient

b-tarczynski:
For now tests are slow as `HttpClient` uses `solidFetch` from common-universal which has `retryDelay` hardcoded. I could change it in this PR, but I am going to drop solidFetch in the next PR anyway


packages/common-nodejs/src/http/HttpClient.test.ts
```diff
@@ -0,0 +1,117 @@
...
+      const httpClient = new HttpClient(Logger.SILENT)
+      await expect(() => httpClient.get(httpServer.getUrl('/status?status=500'), getResponseSchema)).toBeRejectedWith(
+        'Failed GET: 500 - {"status":500}',
+      )
+      expect(httpServer.requestsCount['/status']).toEqual(6)
+    })
+  })
+
+  describe('post', () => {
+    it('returns undefined if no schema provided', async () => {
```
nezouse, line 55: I would say it should either skip validation if schema is undefined or always require schema, this is a little bit weird, not sure why it was like that before

b-tarczynski: It ships the validation [here](https://github.com/sparkdotfi/spark-app/pull/677/files#diff-09d9b9aa01ac4e90b64fd8270877369fbd618f455f704bc63964a58eb8441f8fR31-R32). Is that what you had in mind?

packages/common-nodejs/src/http/HttpServer.mock.ts
```diff
@@ -0,0 +1,61 @@
...
+  public requestsCount: Record<string, number>
+
+  constructor() {
+    this.app = express()
+    this.requestsCount = {}
+    this.setupDefaultRoutes()
+  }
+
+  listen(): void {
+    this.server = this.app.listen(0)
```
krzkaczor, line 18: IMO this deserves a comment that it will find a avaiable port. I didn't know that 0 does that.

packages/common-nodejs/src/http/HttpServer.mock.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    this.app.get('/status', (req, res) => {
+      const status = Number(req.query.status)
+      res.status(status).json({ status })
+    })
+
+    this.app.post('/post', (req, res) => {
+      const bodySafeParse = postBodySchema.safeParse(req.body)
+      if (!bodySafeParse.success) {
+        res.status(404).end()
```
krzkaczor, line 46: It should be `400` for bad request.

packages/common-nodejs/src/http/HttpClient.test.ts
krzkaczor: wtf that's strange. Can we just make schema required and get rid of that?

krzkaczor (commented):
These require more tests for different error code returned from server:
* does retry on 5xx
* does not retry on 4xx (this one should be configurable)

These can be added later once we make implemention better...

packages/common-nodejs/src/http/HttpClient.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  let httpServer: HttpServerMock
+
+  beforeEach(() => {
+    httpServer = new HttpServerMock()
+    httpServer.listen()
+  })
+
+  afterEach(() => httpServer.server?.close())
+
+  describe('get', () => {
```
nezouse, line 17: usually we make it `HttpClient.prototype.get.name` so it's autoupdated on renames

packages/common-nodejs/src/http/HttpClient.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    it('retries in case of server error', async () => {
+      const httpClient = new HttpClient(Logger.SILENT)
+      await expect(() => httpClient.get(httpServer.getUrl('/status?status=500'), getResponseSchema)).toBeRejectedWith(
+        'Failed GET: 500 - {"status":500}',
+      )
+      expect(httpServer.requestsCount['/status']).toEqual(6)
+    })
+  })
+
+  describe('post', () => {
```
nezouse, line 54: same here

nezouse (approved):

packages/common-nodejs/src/http/HttpServer.mock.ts
```diff
@@ -0,0 +1,62 @@
...
+  listen(): void {
+    //finds an available port and binds it to the app
+    this.server = this.app.listen(0)
+  }
+
+  getUrl(path: string): string {
+    const address = this.server?.address()
+    assert(address)
+
+    if (typeof address === 'string') {
```
krzkaczor, line 26: does it even work? can't we simply assert that it's never a string?

krzkaczor (approved):

packages/common-nodejs/src/http/HttpServer.mock.ts
```diff
@@ -21,11 +21,7 @@ export class HttpServerMock {
 
   getUrl(path: string): string {
     const address = this.server?.address()
-    assert(address)
-
-    if (typeof address === 'string') {
-      return address
-    }
+    assert(address && typeof address !== 'string')
```
nezouse, line 24: Shouldn't it assert that it's a string?

PR #678, opened 2025-03-04
Remove trailing comma in network select dialog

yivlad:


oskarvu:
Handled [here](https://github.com/sparkdotfi/spark-app/pull/679)

PR #681, opened 2025-03-04
Add no-extraneous-dependencies config to eslint

pazernykormoran:
closes TS-154
Unfortunately we cannot use biome because necessary options for this rule will be available from biome 2.0 https://biomejs.dev/linter/rules/no-undeclared-dependencies/


packages/common-testnets/package.json
```diff
@@ -51,11 +51,13 @@
...
-    "uuid": "^11.0.2"
+    "uuid": "^11.0.2",
+    "eslint-plugin-import": "^2.31.0"
   },
   "peerDependencies": {
     "viem": "^2.0.0",
     "@marsfoundation/common-nodejs": "workspace:^",
-    "@marsfoundation/common-universal": "workspace:^"
+    "@marsfoundation/common-universal": "workspace:^",
+    "zod": "3.22.4"
```
krzkaczor, line 61: I think should be `^3.0` (all 3 versions are fine)

packages/common-testnets/package.json
```diff
@@ -51,11 +51,13 @@
     "@types/uuid": "^10.0.0",
     "earl": "^1.3.0",
     "mocha": "^10.8.2",
-    "uuid": "^11.0.2"
+    "uuid": "^11.0.2",
+    "eslint-plugin-import": "^2.31.0"
```
krzkaczor, line 55: do we need this in every package.json or maybe just the root one?

pazernykormoran: Basically with this configuration I removed root deps because the rule forces us to have a dependency in each package.

packages/common-nodejs/package.json
```diff
@@ -77,7 +77,8 @@
     "@types/uuid": "^10.0.0",
     "earl": "^1.3.0",
     "express": "^5.0.1",
-    "mocha": "^10.8.2"
+    "mocha": "^10.8.2",
+    "zod": "3.22.4"
```
krzkaczor, line 81: shouldn't zod be a peer dependency then? is it used in runtime or just tests?

pazernykormoran: My bad. Moved to peer deps as its used by HttpClient

krzkaczor (approved):

PR #688, opened 2025-03-05
Improve logs and errors in HttpClient

b-tarczynski:
Requires: https://github.com/sparkdotfi/spark-app/pull/687

### DIFF
--- a/packages/common-nodejs/src/http/HttpClient.ts
+++ b/packages/common-nodejs/src/http/HttpClient.ts
@@ -1,26 +1,34 @@
 import { z } from 'zod'
-
-import { SolidFetch, solidFetch as _solidFetch } from '@marsfoundation/common-universal'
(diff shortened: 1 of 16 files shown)


packages/common-nodejs/src/http/HttpClient.ts
```diff
@@ -26,18 +26,29 @@ export class HttpClient {
...
   }
 }
+
+export class HttpError extends Error {
+  constructor(
+    public readonly method: 'POST' | 'GET',
+    public readonly status: number,
+    public readonly textResult: string,
+  ) {
+    super(`Failed ${method}: ${status} - ${textResult}`)
```
krzkaczor, line 51: IMO this should take url also as a param and make it part of the message.

nezouse (approved):

PR #691, opened 2025-03-06
Move logger to `common-universal`

b-tarczynski:
Additionally we can swap `ms` package with `pretty-ms`.
The differences are:
- `UnixTime.ONE_YEAR()` changes to '1y'
- `UnixTime(UnixTime.ONE_YEAR() - UnixTime.ONE_HOUR())` changes to '364d 23h'

What do you think about swapping `ms` package? It's used only in `UnixTime`


packages/app/vite.config.ts
```diff
@@ -28,7 +28,8 @@ export default defineConfig({
     // nodePolyfills needs to be first
     nodePolyfills({
       // buffer is needed when connecting with a coinbase wallet installed on a phone
-      include: ['buffer'],
+      // util and path is needed for common-universal's logger
+      include: ['buffer', 'util', 'path'],
```
nezouse, line 32: Do we really need these? I would prefer not to use node apis in browser

b-tarczynski: `Util` is used for `inspect` to format objects, path isn't doing anything in browser. I tried to conditionally import a path depending if that's a browser or not, but nothing worked - logger would have to be async which we don't want to do

b-tarczynski: Anyway using `util` polyfill breaks frontend tests, so probably we need to figure out something else

nezouse (approved):

packages/common-universal/src/logger/Logger.ts
```diff
@@ -38,7 +34,7 @@ export class Logger implements ILogger {
       service: options.service,
       tag: options.tag,
       utc: options.utc ?? false,
-      cwd: options.cwd ?? process.cwd(),
+      cwd: options.cwd,
```
krzkaczor, line 37: Is it possible to use process when not in browser?

b-tarczynski: Before we were using node:path's `join` function after obtaining cwd, that's why I haven't done this in common-universal

b-tarczynski:
Now we have Logger.BROWSER that would console.log the object with logger information and parameters that could be viewed in browser's console

packages/common-universal/src/logger/LogFormatterObject.ts
```diff
@@ -0,0 +1,15 @@
+import { LogEntry, LogFormatter } from './types.js'
+
+export class LogFormatterObject implements LogFormatter {
```
krzkaczor, line 3: I think it's so similiar to json formatter that we can drop it 👍

krzkaczor (approved):

PR #696, opened 2025-03-07
Move httpClient to common-universal

b-tarczynski:


packages/common-universal/src/httpClient/HttpClient.test.ts
```diff
@@ -1,8 +1,8 @@
-import { Logger } from '@marsfoundation/common-universal/logger'
 import { MockObject, expect, mockFn, mockObject } from 'earl'
 import { ZodError, z } from 'zod'
+import { Logger } from '../logger/Logger.js'
 import { HttpClient, HttpError } from './HttpClient.js'
-import { HttpServerMock, PostBody, getResponseSchema, postBodySchema } from './HttpServer.mock.js'
+import { HttpServerMock, PostBody, getResponseSchema, postBodySchema } from './HttpServer.mock'
```
nezouse, line 5: Why extension was removed here?

b-tarczynski: Added back

krzkaczor: linter should generally catch it... even though in tests it doesnt matter that much.

packages/common-universal/package.json
```diff
@@ -26,6 +26,18 @@
         "default": "./dist/cjs/index.js"
       }
     },
+    "./httpClient": {
```
krzkaczor, line 29: i wonder shouldn't it be `http-client` as it's package name part 🤔

packages/common-universal/src/httpClient/HttpClient.test.ts
krzkaczor: can you verify biome configuration?

krzkaczor (approved):

PR #705, opened 2025-03-10
Filter out rewards for blocked countries

oskarvu:
Closes FRO-1376


packages/app/src/domain/spark-rewards/claimableRewardsQueryOptions.ts
```diff
@@ -39,7 +41,9 @@ export function claimableRewardsQueryOptions({
       ? skipToken
       : async () =>
           (
-            await Promise.all(sparkRewardChainIds.map((chainId) => fetchChainRewards(chainId, wagmiConfig, account)))
+            await Promise.all(
+              sparkRewardChainIds.map((chainId) => fetchChainRewards({ chainId, wagmiConfig, account, countryCode })),
```
yivlad, line 45: I don't think it's a good idea to include country code into query fn. Right now I think this wouldn't work for rewards on mainnet after refresh because when the query fn first starts the country code is undefined. I think this should be aggregated latter - after both query succeed.

oskarvu: Good point!

yivlad (approved):

PR #723, opened 2025-03-13
Track savings deposits as separate events

oskarvu:
Closes FRO-1389

### DIFF
--- a/packages/app/src/features/actions/flavours/deposit-to-savings/logic/depositToSavingsAction.ts
+++ b/packages/app/src/features/actions/flavours/deposit-to-savings/logic/depositToSavingsAction.ts
@@ -10,6 +10,7 @@ import {
   usdcVaultAbi,
   usdsPsmActionsConfig,
 } from '@/config/contracts-generated'
+import { trackEvent } from '@/domain/analytics/mixpanel'
 import { getContractAddress } from '@/domain/hooks/useContractAddress'
 import { ensureConfigTypes } from '@/domain/hooks/useWrite'
 import { EPOCH_LENGTH } from '@/domain/market-info/consts'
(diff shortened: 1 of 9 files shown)


packages/app/src/features/actions/flavours/deposit-to-savings/logic/depositToSavingsAction.ts
```diff
@@ -231,6 +232,17 @@ export function createDepositToSavingsActionConfig(
           assertNever(actionPath)
       }
     },
+
+    onSuccessfulAction: () => {
```
yivlad, line 236: Wouldn't this mean that this event would be tracked twice? Here and in useWrite hook.

I'm thinking of an alternative, maybe we should use the same patter as in `useIncreasedGas` hook to determine what transaction actually does and based on this do something like
```ts
const finalWrite =
    parameters && enabled
      ? () => {
          sanityCheckTx(parameters.request, chainId)
          callbacks.onBeforeWrite?.()

          const sentryTransactionType = getSentryTransactionType()
          
          if (sentryTransactionType === 'deposit-to-savings') {
            trackEvent('deposit-to-savings-tx-sent', {
              walletType,
              chainId,
              receiver: parameters.request.address,
              method: parameters.request.functionName,
              args: JSONStringifyRich(parameters.request.args),
            })
            } else {
              trackEvent('tx-sent', {
              walletType,
              chainId,
              receiver: parameters.request.address,
              method: parameters.request.functionName,
              args: JSONStringifyRich(parameters.request.args),
            })
            }
  
            writeContract(parameters.request as any)
          }
      : () => {}
```

yivlad (approved):

packages/app/src/features/actions/flavours/deposit-to-savings/logic/depositToSavingsAction.ts
oskarvu: I would leave tx-sent as a generic event, which is still useful to have, even if more specific events will be send from savings.

Having specific event for each deposit, like: `base-usdc-to-susdc` or `arbitrum-usdc-to-susdc` enables more sane way to crate data sets. When sending from here, we are loosing a lot of context, e.g., we can't easily get amount, cause different deposit methods has different args. We don't have access to token formatting methods and so on.

PR #725, opened 2025-03-13
Update rewards endpoints with network parameters

yivlad:


packages/app/src/domain/spark-rewards/claimableRewardsQueryOptions.ts
```diff
@@ -74,7 +74,7 @@ async function fetchChainRewards(
   })
 
   // @todo: Add chainId parameter to the API call
```
oskarvu, line 76: Comment can be removed.

packages/app/src/features/market-details/logic/useSparkRewards.ts
```diff
@@ -15,7 +15,7 @@ export function useSparkRewards({ chainId, reserve }: UseSparkRewardsParams): Us
   const ongoingCampaignsResult = useOngoingCampaignsQuery()
 
   const { data } = transformSimplifiedQueryResult(ongoingCampaignsResult, (data) => {
-    const campaigns = data.filter((campaign) => campaign.chainId === chainId)
+    const campaigns = data.filter((campaign) => campaign.type === 'sparklend' && campaign.chainId === chainId)
```
oskarvu, line 18: Filtering by campaign type is also made in `assignMarketSparkRewards`, so maybe we should change this variable to `sparklendCampaigns`, and in `assignMarketSparkRewards` accept `sparklendCampaigns` instead of `campaigns`. But I think filtering it later in `assignMarketSparkRewards` (same as it works now) is also ok.

oskarvu (approved):

PR #731, opened 2025-03-15
Add cookie3 analytics

yivlad:


packages/app/index.html
```diff
@@ -30,6 +30,18 @@
     <meta name="keywords"
       content="Decentralized Finance, DeFi, lending, borrowing, stablecoins, Ethereum, assets, erc-20, smart contracts, open finance, trustless" />
     <script defer data-domain="%VITE_ANALYTICS_PLAUSIBLE_ID%" src="https://plausible.io/js/script.js"></script>
+    <script>
+      const cookie3Options = {"siteId": '%VITE_ANALYTICS_COOKIE3_ID%',"additionalTracking":true,"cookielessEnabled":true};
+
+      window._paq = window._paq || [];
+      (function () {
+          var d = document, g = d.createElement('script'), s = d.getElementsByTagName('script')[0];
+          g.async = true; g.src = 'https://cdn.cookie3.co/scripts/analytics/latest/cookie3.analytics.min.js';
```
oskarvu, line 39: What scares me a bit is that we are using cdn with "latest" tag. @krzkaczor What do you think? Maybe we should ask cookie3 people if there is a way to fetch specific version?

krzkaczor: @oskarvu yes, very good point. Maybe we can simply download the current version and serve it from static? (there is also npm package so whatever is easier).

oskarvu: Yep, the easiest will be just download and serve it static. Plugin version is very similar in functionality with mixpanel, so I think we should just use script.

We should probably also ask clay to serve script statically, if they don't do it already.

oskarvu (approved):

PR #746, opened 2025-03-20
Correctly render new lines in Slack rendering

b-tarczynski:
Previously we couldn't render new lines on it own and new lines would be always rendered at the end of block. Tests that I added show how it suppose to work

Closes: TS-273


nezouse (commented):
Wont it break if you want to have ie text [link] text in the same line? It will be always split by new lines, right?

packages/common-reporters/src/slack/renderToSlack.test.ts
```diff
@@ -28,6 +28,44 @@ describe(renderToSlackString.name, () => {
...
+      templating.text('Fourth line'),
+    ]
+
+    const lines = ['> First line', '> ', '> Third line', '> Fourth line']
+    expect(renderToSlackString(content)).toEqual(lines.join('\n'))
+  })
+
+  it('renders new lines in correct order', () => {
+    const content = [
+      templating.text('First'),
```
krzkaczor, line 58: how about aliasing tempalating to t in this file :P

krzkaczor (approved):

PR #750, opened 2025-03-21
Remove frontend's unused dependencies

b-tarczynski:
Closes: TS-280


packages/app/package.json
```diff
@@ -64,9 +63,6 @@
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.0.0",
     "d3-array": "^3.2.4",
-    "deepmerge-ts": "^5.1.0",
-    "fetch-retry": "^5.0.6",
-    "jsbi": "^3.2.5",
```
b-tarczynski, line 69: Is it used? I don't see any usages and in docs they say that it's need to be imported and used in order to work

PR #751, opened 2025-03-21
Add claim spark unstake action

yivlad:


packages/app/src/features/actions/flavours/claim-unstake-spark/logic/claimUnstakeSparkAction.ts
```diff
@@ -0,0 +1,32 @@
...
+        address: vault,
+        abi: testSparkStakingConfig.abi,
+        functionName: 'claimBatch',
+        args: [account, epochs.map(BigInt)],
+      })
+    },
+
+    invalidates: () => {
+      return [
+        // @todo: spark staking - add related query keys after they are implemented
```
oskarvu, line 28: In this case balances will for sure change, so we can add it already.

oskarvu (approved):

PR #755, opened 2025-03-24
Assemble SPK view

yivlad:


packages/app/src/features/spk-staking/views/SpkStakingView.stories.ts
```diff
@@ -0,0 +1,94 @@
...
+    },
+  },
+}
+
+export default meta
+type Story = StoryObj<typeof SpkStakingView>
+
+export const Desktop: Story = {}
+export const Mobile = getMobileStory(Desktop)
+export const Tablet = getTabletStory(Desktop)
```
oskarvu, line 75: Let's go with one column on tablet. Please, add also mobile and tablet stories for active variant.

packages/app/src/features/spk-staking/components/page-header/PageHeader.tsx
oskarvu, line 1: I would inline it tbh, as its used only in one place. We will be suspending view, so it won't be used later for loading/error views.

oskarvu (approved):

PR #762, opened 2025-03-26
Add Stake SPK dialog

oskarvu:
Closes FRO-1413


packages/app/src/features/spk-staking/dialogs/stake/components/TransactionOverview.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <TransactionOverview>
+      <TransactionOverview.Row>
+        <TransactionOverview.Label>Estimated rewards</TransactionOverview.Label>
+        <TransactionOverview.FarmApy apy={apy} rewardsPerYear={rewardsPerYearUsd} rewardToken={usds} />
+      </TransactionOverview.Row>
+      <TransactionOverview.Row>
+        <TransactionOverview.Label>Unstaking delay</TransactionOverview.Label>
+        <TransactionOverview.Generic>
+          {UnixTime.formatSpan(UnixTime(unstakingDelay), { unitCount: 2 })}
```
yivlad, line 39: Maybe should reuse logic from `TimeLeftCell`. `formatSpan` couldn't be used there because 0 span was being formatted weirdly. Here this is not the case, but maybe we should do this to be consistent

packages/app/src/features/spk-staking/dialogs/stake/StakeDialogContentContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { withSuspense } from '@/ui/utils/withSuspense'
+import { Percentage } from '@sparkdotfi/common-universal'
+import { SpkStakingEpochs } from '../../types'
+import { useStakeDialog } from './logic/useStakeDialog'
+import { StakeView } from './views/StakeView'
+
+export interface StakeContainerProps {
+  closeDialog: () => void
+  apy: Percentage
+  spkStakingEpochs: SpkStakingEpochs
```
yivlad, line 12: Why does it have to be parametrized this way and not fetched inside the master hook?

yivlad: Can be a placeholder for now

packages/app/src/features/spk-staking/dialogs/stake/views/StakeView.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+    },
+  } satisfies Omit<StakeViewProps, 'form'>,
+}
+
+export default meta
+type Story = StoryObj<typeof StakeView>
+
+export const Desktop: Story = {}
+export const Mobile = getMobileStory(Desktop)
+export const Tablet = getTabletStory(Desktop)
```
yivlad, line 59: Add empty input story

packages/app/src/features/spk-staking/dialogs/stake/logic/useStakeDialog.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      epochDuration: UnixTime(604800),
+      epochDurationInit: UnixTime(1737071795),
+    },
+  },
+})
+
+describe(useStakeDialog.name, () => {
+  beforeEach(() => {
+    vi.useFakeTimers({
+      shouldAdvanceTime: true,
```
yivlad, line 35: Why?

oskarvu: Didn't work without it. Even if I manually progressed time. I have deterministic tests with auto progressing.

packages/app/src/features/spk-staking/dialogs/stake/StakeDialogContentContainer.tsx
oskarvu: I didn't want to make similar query twice, since we will need this data also inside staking view. But no strong opinion here. Can also call from dialog hook. The only thing is that we will be fetching it twice, since both dialog and main view queries are suspended and have different queries.

yivlad (approved):

PR #765, opened 2025-03-26
Improve `ZodString` detection

b-tarczynski:
For some reason `instanceof` check didn't work in dns-sentinel even though I was using the same Zod version. I added additional check for `typeName` which solved the problem

### DIFF
--- a/packages/common-universal/src/httpClient/HttpClient.ts
+++ b/packages/common-universal/src/httpClient/HttpClient.ts
@@ -29,7 +29,7 @@ export class HttpClient {
       throw new HttpError('POST', url, result.status, await result.text())
     }
 
-    if (schema instanceof ZodString) {
+    if (isZodString(schema)) {
       return await result.text()
     }
     return schema.parse(await result.json())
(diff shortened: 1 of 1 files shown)


packages/common-universal/src/httpClient/HttpClient.ts
```diff
@@ -67,3 +67,7 @@ export class HttpError extends Error {
 const applicationJsonHeader = {
   'Content-Type': 'application/json',
 }
+
+function isZodString<T extends z.ZodTypeAny>(schema: T): boolean {
+  return schema instanceof ZodString || schema._def.typeName === 'ZodString'
```
nezouse, line 72: What about zod.instanceof? Maybe it will work better

krzkaczor: `z.instanceof` is a validator i think. I doubt it will work any better. IMO this looks fine. 

One idea to improve, it could be a guard function.

krzkaczor (approved):

